PR #3, opened 2020-12-17
Remove extra pull in example docker pull update command

zPaw:
Just a simple README.md update


meeb:
You are technically correct. The best kind of correct. Thanks!

PR #29, opened 2020-12-30
Added format descriptions.

ltomes:
One of my first questions while evaluating tubesync was what format options are available. 
I added format descriptions to the media format help_text based on the options I see in the models.

Edit: I am not sure why the last pipeline is failing, My changes should not relate to that.

### DIFF
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -200,7 +200,28 @@ class Source(models.Model):
         _('media format'),
         max_length=200,
         default=settings.MEDIA_FORMATSTR_DEFAULT,
-        help_text=_('File format to use for saving files')
+        help_text=_("""File format to use for saving files
+Valid options:
+    'yyyymmdd': 20210201,
+    'yyyy_mm_dd': 2021-02-01,
+    'yyyy': 2021,
+    'mm': 02,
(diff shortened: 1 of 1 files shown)


meeb:
Hi, thanks for the contribution. Yeah detailing the format options is pretty important. I'm not entirely sure abusing the `help_text` attr with a big block of info works too well for the UI and messes up the auto-generated forms quite a lot. You might have missed it as it's kind of hidden but there is a big table of all the format options at the bottom of the create and update source forms, under the button. This one:

https://github.com/meeb/tubesync/blob/main/tubesync/sync/templates/sync/_mediaformatvars.html

Which looks like this:

Screenshot from 2020-12-30 15-07-35
... etc. options truncated.

Does this not do what you want? I'd rather direct people to the existing table of options where possible.

ltomes:
Oh wow, yes I just missed that on the page.

I added a note to the end of the helper text if that seems useful; but I suspect _most_ people found those options right away and I'm an edge case. Feel free to close this pr!

meeb:
I suspected people may miss it when I was stuffing it at the end of the page but I couldn't really think of a better place to add it, your PR obviously shows people did miss it so the extra note helps. Thanks!

Issue #90, opened 2021-03-05
Weird behavior adding multiple playlists with shared video(s)


micahmo:
Hi @meeb, I'm back again. :-) I think I discovered a very interesting issue related to adding multiple playlists that contain some of the same videos. It just so happened that I was adding two playlists from the same channel which shared a video, and there were some interesting consequences. (Actually, I wanted to add three playlists from the same channel where A and B shared a video, and B and C shared a different video. But two is enough to repro the issue.)

Here's how you can reproduce. The settings for the source don't seem to matter.
 - Add this playlist as a source: https://www.youtube.com/playlist?list=PLaDrN74SfdT6nO7cnR0zcfM6Bc5UQuKXG
   - This playlist has 4 videos. TubeSync correctly identifies 4 media items and downloads them all. So far so good.
 - Add this playlist as a source: https://www.youtube.com/playlist?list=PLaDrN74SfdT5jEs3RCI53nBUkuURSWrhq
   - This playlist has 25 videos and shares one video (oCp_2hqguUg) with the first playlist.
   - TubeSync correctly identifies the 25 media items.

Now there are two interesting problems with the system.
 - TubeSync only downloads 24 videos from the second playlist. It skips the shared video, presumably because it thinks it is already downloaded.
 - TubeSync now lists only 3 media items for the first playlist (this happens immediately after adding the second).

So the shared video is downloaded for one playlist but not listed in its media items, and it's not downloaded for the other playlist where it is listed in the media items.

Let me know if there's anything else I can do to test for you. And sorry to put more work on your plate, but I just wanted to report the issue. I'm only finding these interesting scenarios because I am using TubeSync a lot. :-)

meeb:
Thanks for the detailed description, I'll have a go at replicating this. It's quite possible there's a bug in handling of media items as unique per (source, video_id) and just using the video_id in some places which would cause this. I'll log this as a bug for now.

PhuriousGeorge:
I've got some channels I monitor both audio and video from.  I've added basically duplicate sources and changed the KEY value to include '-music' on the duplicate entry.   I hadn't thought about checking this scenario until reading the issue above, but another scenario that might be affected.  I'll have a go at adding them and see what comes of it.

meeb:
[...]

PR #95, opened 2021-03-12
Fix meeb/tubesync#90: Match media on source and id

micahmo:
When looking for existing media objets during source indexing, the query should match both the key (the video's ID) and the source. Otherwise, a media object with a given key can only ever exist once, which breaks the case where multiple playlists share a video.

### DIFF
--- a/tubesync/sync/tasks.py
+++ b/tubesync/sync/tasks.py
@@ -175,7 +175,7 @@ def index_source_task(source_id):
             # Video has no unique key (ID), it can't be indexed
             continue
         try:
-            media = Media.objects.get(key=key)
+            media = Media.objects.get(key=key, source=source)
         except Media.DoesNotExist:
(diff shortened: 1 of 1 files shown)


micahmo:
Fixes meeb/tubesync#90 (adding comment for linkage)

meeb:
All looks fine, thanks! I believe possibly (and maybe incorrectly off the top of my head) that there may be additional places this check is performed as well. However your PR does indeed fix one of the more obvious problems where I fat-fingered a query or two. I'll leave the original issue open for now as I think this whole matching code is a bit hacky and needs a larger review which is why it's been open for a while. Thanks again, much appreciated.

micahmo:
Sounds good! Just for awareness, the merge closed the issue automatically, so you'll have to reopen it if you want to review it some more (but it's resolved as far as I'm concerned).

Issue #88, opened 2021-03-05
[Enhancement] More control over fallback selection


micahmo:
Hi, first of all I am loving tubesync. I've had in mind for a while that I wanted to automatically download playlists for viewing in Plex, and I was so glad to discover that someone has already done the hard work to develop it. :-) (BTW, I just stumbled upon this project when I was browsing the New Apps section of the Community Apps in Unraid.)

It would be nice if there were some additional options related to the selection of the fallback quality and/or codec.

Here is my use case. I set up a playlist source where the source resolution is set to `1080p (Full HD)`, the source video codec is set to `AVC1 (H.264)` or `VP9` (I experimented with both), and the fallback is set to `Get next best resolution but at least HD`.

I am downloading a playlist where almost all of the videos are 1080p with both AVC1 and VP9 codec options available, which is great. However, there's a single video that's only 720p. Since it doesn't have a 1080p option, the fallback logic kicks in. I'm not sure what the logic is, but my guess is that it does something like `bestvideo+bestaudio` in youtube-dl. Unfortunately, the `bestvideo` for this video uses AV1 codec, which [Plex does not yet support](https://forums.plex.tv/t/add-support-for-av1-coding-standard/556253). There are plenty of other codecs available at 720p which would work perfectly fine, but the fallback options aren't detailed enough for me to pick, so I'm stuck with the AV1 unless I download the video manually.

If you're curious, here is the video ID: GSOo84fAdsE. Running `youtube-dl -F GSOo84fAdsE | grep 720` gives these formats. I would be happy with any of them except 398, which unfortunately happens to be the best in terms of bitrate. :-)

```
136          mp4        1280x712   720p  447k , mp4_dash container, avc1.4d401f@ 447k, 30fps, video only, 112.42MiB
247          webm       1280x712   720p  551k , webm_dash container, vp9@ 551k, 30fps, video only, 138.35MiB
398          mp4        1280x712   720p  647k , mp4_dash container, av01.0.05M.08@ 647k, 30fps, video only, 162.58MiB
22           mp4        1280x712   720p  576k , avc1.64001F, 30fps, mp4a.40.2 (44100Hz) (best)
```

I'll leave it up to you to decide if/how you want to handle this case. Here are a couple of ideas I had.

You could add the ability to blacklist a particular format. For me, I know that I will never want AV1 since it's not supported by Plex.

[...]

PR #100, opened 2021-03-17
Related to #88: Fix 4320p in test data.

micahmo:
Related to #88: Fix 4320p in test data.

Just a small fix to the test data. The tests pass with or without the change, but I discovered that `RESOLUTION_MAP` has the key `SOURCE_RESOLUTION_4320P`, which is case-sensitive with a lower-case `p`. If future changes are made that cause the tests to fail, this fix will eliminate one potential reason for failure due to bad data.


meeb:
Thanks! Pretty sure this is run through `.upper()` or something before checking anyway but every little tweak helps :)

Issue #202, opened 2022-01-14
Generate metadata as json


bullyrooks:
Hi,  I'm trying to integrate tubesync with plex.  As you are probably aware, indexing youtube data in plex is pretty awful.  I was planning on using this tool:
https://github.com/JordyAlkema/Youtube-DL-Agent.bundle
which uses the json file generated (via flag) by youtube-dl as the input into the scanner/agent for indexing.  I _think_ youtube-dlp has similar functionality (their docs say they do)
https://github.com/yt-dlp/yt-dlp#filesystem-options

would it be possible to update the store functionality to include an option to emit the metadata json file, similar to the .nfo function?  This would really help me out.  Otherwise, if there's a better or alternate solution, I would love to hear it. 

I'm willing to write up some docs on my blog (https://bullyrooks.com/index.php/category/technology/home-networking/media/) which (hopefully) could help others get this working as well.

Appreciate your work.  Thanks!

meeb:
There appears to be an embedded API call to write the metadata as a `.info.json` file. I'll give it a test to see if it does what you suggest. I have no issue with adding this feature.

bullyrooks:
awesome.   I'm very interested in that.  If you get a chance to implement it, let me know and I'll pull down, redeploy, test and write up some docs.  Appreciate it!

mcinj:
Just to see if it was possible, [I added parsing of Tubesync's .nfo xml](https://github.com/mcinj/Youtube-DL-Agent.bundle/tree/tubesync_xml) to Youtube-DL-Agent.bundle. It does work as long as the media, nfo, and image file have the same name. #204 will have to get resolved for it to work automatically for either .json or .nfo file as https://github.com/JordyAlkema/Youtube-DL-Agent.bundle also currently looks for a matching .json and image file with the same name, just different extension.

ThibaultNocchi:
Hello,

I am too interested in this feature as the YT metadata plugin for Jellyfin supports reading them from the JSON generated by yt-dl. https://github.com/ankenyr/jellyfin-youtube-metadata-plugin

Thanks!

meeb:
This issue should be resolved thanks to the lovely work by @ThibaultNocchi

go2tom42:
[...]

PR #211, opened 2022-02-01
Add possibilty to use yt dlp write-info-json flag

ThibaultNocchi:
Hello,

As I described [here](https://github.com/meeb/tubesync/issues/202#issuecomment-1027029633), I am very much interested into having the possibility to generate the JSON file from the '--write-info-json' flag of yt-dlp. This would help me use the Jellyfin extension for YT metadata as it can use those files.

Here is the PR for it.

* Added a boolean field for Source to enable write_json
* Added Media properties to get the path of the JSON file
* Added the HTML field in the template
* Added the field to the create and update views
* Added the delete_file method for the JSON when deleting a source
* Add the flag in the download function

Please tell me if there's something missing, I'll try my best to suit your needs. :)


meeb:
This looks great, thanks for the PR! I'll merge it and roll it into the next release.

PR #246, opened 2022-06-08
Build docker image for amd64 and arm64

SweetMNM:
As per the title. You should note the build time for arm64 takes a while (entire containerize job was 33 minutes) because python needs to build the libsass library for arm64 every time. libsass does not provide arm64 prebuild neither does piwheels. I am not very familiar with python, is there a for us the prebuild libsass once and have it download that?

I tested the image on my PI 4 and it works perfectly, if anyone wants to test or use it for now, you can use:
`ghcr.io/sweetmnm/tubesync`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,19 +35,24 @@ jobs:
   containerise:
     runs-on: ubuntu-latest
     steps:
-    - name: Set up QEMU
-      uses: docker/setup-qemu-action@v1
-    - name: Set up Docker Buildx
-      uses: docker/setup-buildx-action@v1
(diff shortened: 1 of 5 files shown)


meeb:
This looks fantastic, thanks so much for the contribution. I'll give this a test shortly. Just to confirm this builds a multi-arch container image so `:latest` works on both amd64 and arm64?

SweetMNM:
Correct! You can see [here](https://github.com/users/sweetmnm/packages/container/tubesync/24554337?tag=latest)

meeb:
Looks pretty good so far! Merged, will test further. Thanks again.

Issue #260, opened 2022-08-06
To many thumbnails


Mrpuppyz:
Is there a way to have tubesync only download thumbnails for videos I am actually downloading? I have it set to only download videos from the last week however it still downloads every single thumbnail which I do not need. I only wanted to download the thumbnails for the videos from the last 7 days or videos actually downloading. If this cannot be done is there a way to completely disable downloading the thumbnails all together? I really only need the videos thumbnails are not that big of a deal for me.

meeb:
There isn't currently, but that's certainly a sensible idea. I'll add it to the wishlist.

MattTheEmp:
Meeb first off love the project but this issue really needs to be addressed. It will remove much overhead processing from your application and save data.

keithcroshaw:
Trying this all again as I've been playing with ytdl-sub-gui and missed the nice UI. It takes so long to get all the metadata before it can even get started. Is it getting metadata for every video on the channel? That hurts... 🤣

meeb:
Yeah, the full metadata for each media item is needed (even if it's just for the upload date to use as the download cutoff). There isn't a better way to handle it unfortunately. It's only once though on first adding a channel, it's not refreshed every crawl or anything.

Mrpuppyz:
After I added an external database to my set up. It doesn't seem to take as long for the first index. I am pretty much content with the way it works now. Thank you for building this awesome project.

meeb:
Thanks! And yeah, if you want to index big channels an external database is certainly advisable. For anyone else stumbling over this issue:

https://github.com/meeb/tubesync/blob/main/docs/other-database-backends.md

PR #293, opened 2022-12-08
Skip metadata and thumbnails if they are past the cutoff date

geoah:
Hey, I'm playing around with tubesync and I'm trying to add a couple of large channels and bumped into the #260 issue.

Would something like this work? I'm not really familiar with django or the process that media and tasks go through so I might be very off base here. I am also making the assumption that media.skip has already been set on the media entry before we get to the signaling for the thumbnails, which might not be correct.

I'm currently having some issues building the container or even installing deps locally on macos so I haven't tested this; I blame the fact it's pretty late here and will give it another shot tomorrow.

Any issues and ideally suggestions for getting this right would be appreciated!
Regards, George

### DIFF
--- a/tubesync/sync/signals.py
+++ b/tubesync/sync/signals.py
@@ -147,7 +147,7 @@ def media_post_save(sender, instance, created, **kwargs):
         instance.save()
         post_save.connect(media_post_save, sender=Media)
     # If the media is missing metadata schedule it to be downloaded
-    if not instance.metadata:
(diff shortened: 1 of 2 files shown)


meeb:
This is fine for the thumbnail skipping, however I'll revert the changes to signals.py as these break the ability to refresh metadata. This finally makes a some sense to merge, so thanks for the PR even if it took over a year to merge!

PR #306, opened 2023-01-17
load youtube channels from www.youtube.com/{key}

biolds:
Some channel return a 404 with the /c/ prefix
(ie. https://www.youtube.com/c/@relaxingmusicandvibes)


biolds:
Follow up on https://github.com/meeb/tubesync/pull/305#issuecomment-1383366647, how should these kind of channels be supported ? I didn't know there were different kind of channels..

meeb:
That's because they're not channels, they're user pages or other aliases. The only real way to make a sane interface for people (probably)  is to just accept any URL, pull the metadata via `yt-dlp` and then check for a channel ID, then suggest using that. For example, the actual channel name for [@relaxingmusicandvibes](https://www.youtube.com/c/@relaxingmusicandvibes) is https://www.youtube.com/channel/UCxbIu0sL2NwqzbeU65CxAbw which isn't exactly obvious.

plaes:
Another example of one of these channels - https://www.youtube.com/@ShoyanJapaneseCarpenter

It's actually possible to subscribe to this "page" separately.

PR #320, opened 2023-02-03
feat: initial sponsorblock support

ticoombs:
- Provides inbuilt sponsorblock support via the postprocessor
- Writes sponsorblock as the chapter file
- No removal of segments or configuration options currently (I'm not skilled enough in Django)


meeb:
Excellent, thanks! I'll tweak this slightly to make it an opt-in flag per source before releasing it. Thanks for the PR.

PR #325, opened 2023-02-11
Fix bug getting content_type for "audio only"

kuhnchris:
Content type was throwing error when `video_type` was set to "Audio only". 
This rectifies this, not 100% sure if it's better, but it's better than a dump.
I assume there is a `downloaded_video_codec` aswell, we could use that and split it accordingly (OPUS/VP9) - let me know how you'd prefer to handle this.

### DIFF
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -1140,7 +1140,11 @@ class Media(models.Model):
     def content_type(self):
         if not self.downloaded:
             return 'video/mp4'
-        vcodec = self.downloaded_video_codec.lower()
+        vcodec = self.downloaded_video_codec
+        if vcodec is None:
+            return 'audio/ogg'
+
+        vcodec = vcodec.lower()
         if vcodec == 'vp9':
             return 'video/webm'
         else:
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -1142,7 +1142,18 @@ class Media(models.Model):
             return 'video/mp4'
         vcodec = self.downloaded_video_codec
         if vcodec is None:
-            return 'audio/ogg'
+            acodec = self.downloaded_audio_codec
+            if acodec is None:
+                raise TypeError() # nothing here.
+            
+            acodec = acodec.lower()
+            if acodec == "mp4a":
+                return "audio/mp4"
+            elif acodec == "opus":
+                return "audio/opus"
+            else:
+                # fall-fall-back.
+                return 'audio/ogg'
 
         vcodec = vcodec.lower()
         if vcodec == 'vp9':


meeb:
It would probably be more sensible to use the `Media.downloaded_audio_code` database field if the `vcodec` is None, there's no guarantee elsewhere that the audio format is always going to be `audio/ogg`. Would you mind resubmitting this using the `downloaded_audio_code` field? If not I can tweak it before the next release. Thanks for the PR.

kuhnchris:
Sure, I'll adapt it accordingly and add a commit to this branch, let me know if I shall squash it or if it's fine to have 2 commits for it.
🥂

kuhnchris:
```kuhnchris@ERIKA-MINIPC:/mnt/c/Users/kuhnc$ curl -I http://127.0.0.1:8000/media-content/0ea6c203-2835-47d5-992c-7dc884990d08
HTTP/1.1 200 OK
Date: Sun, 12 Feb 2023 12:21:39 GMT
Server: WSGIServer/0.2 CPython/3.10.6
Content-Type: video/webm
X-Accel-Redirect: /media-data/home/kuhnchris/tubesync/tubesync/downloads/2022-01-12_daaaaaaada_youve-reached-your-limit-on-talking_qpWJ5R_hU6I_1080p-vp9-opus-60fps.mkv
X-Frame-Options: SAMEORIGIN
Content-Length: 0
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin

kuhnchris@ERIKA-MINIPC:/mnt/c/Users/kuhnc$ curl -I  http://127.0.0.1:8000/media-content/6040f6a8-a4be-49db-9b5c-4c46533e8524
HTTP/1.1 200 OK
Date: Sun, 12 Feb 2023 12:22:06 GMT
Server: WSGIServer/0.2 CPython/3.10.6
Content-Type: audio/opus
X-Accel-Redirect: /media-data/audio/feee/2016-11-14_aaaaaaaaaaaaa_op_CquikyW3PPs_opus.ogg
X-Frame-Options: SAMEORIGIN
Content-Length: 0
X-Content-Type-Options: nosniff
Referrer-Policy: same-origin
```

*technically* I'm not happy that `vp9` in a `mkv` is sent as `video/webm`, as it is actually a container format (`video/x-matroska`) - but if we are *that* technical we might aswell use `python-magic` to determine the "real" magic type anyways, and I'm not sure if that may not be overkill.

meeb:
Yep much more flexible, thanks. I agree with you on the technicalities, but there's only so much you can do when wrapping formats in containers. This is likely sufficient for almost all cases (and a significant improvement on just assuming all audio is vorbis/ogg). Thanks!

PR #326, opened 2023-02-11
<audio>-player instead of <video> for audio only

kuhnchris:
Strictly speaking you *can* use the `<video>` player for `<audio>`, but it creates a huge gap and makes preview pretty badly usable IMHO.

image

instead of:

image

(yes, this is a cosmetic change; low prio)


meeb:
Nice, thanks!

PR #327, opened 2023-02-11
(dev only) allow file download directly from django

kuhnchris:
```This is a development change only - no changes on prod!```
... also requires setting `PROVIDE_FILES_DIRECTLY = True` in `local_settings.py` .

With this change, the entire package can be self-contained to test and develop without using a `docker` or `devcontainers` container.

### DIFF
--- a/tubesync/sync/views.py
+++ b/tubesync/sync/views.py
@@ -1,8 +1,10 @@
 import os
 import json
 from base64 import b64decode
+import pathlib
+import sys
 from django.conf import settings
-from django.http import FileResponse, Http404
+from django.http import FileResponse, Http404, HttpResponseNotFound
 from django.views.generic import TemplateView, ListView, DetailView
 from django.views.generic.edit import (FormView, FormMixin, CreateView, UpdateView,
                                        DeleteView)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by meeb** [comment]
You could probably just use `settings.DEBUG` along with a check for the Django development server for this rather than introduce another toggle that can be set separately from `DEBUG`. Given this feature is only going to be used in situations where the Django dev server is being used and also not behind an upstream nginx server you could do something like:

```python
if settings.DEBUG and 'runserver' in sys.argv:
   # use manual FileResponse(...) download
else:
  # use nginx X-Accel-Redirect headers
```

as the feature toggle? Also if possible this should use `pathlib` and also contain a check for `.is_file()`, raising a 404 if the file doesn't exist before calling `FileResponse(...)`. Thanks for the PR!

kuhnchris:
Sure, I'll adapt accordingly and let you know if it's ready!

kuhnchris:
Added check, also added pathlib for the final path, also changed the trigger to Debug + "runserver".
Let me know if that works for you. :-)

meeb:
Much nicer, thanks!

PR #331, opened 2023-02-12
"Sync now" button

kuhnchris:
This PR adds a nifty litte "Sync" button that schedules a reindex.
image

I did *not* create a new View(class) for this, I only split the get() path into "sync" and "not sync".
I hope this is fine by you  @meeb ? If not please let me know and I will try to create a new View for it.

Also, this fixes a weird bug that kept showing that a task is "reoccuring" alltho it's "repeat" property is set to 0 in the task overview. Let me know if I shall split that into 2 PRs.

Also includes the additional "arrow-rotate-right" from FontAwesome, for some reason that one wasn't in the SCSS? 🤔 
It's not a "PRO" icon, so I'm just a bit confused if this is just a `lite` version of FA.

Also, this changes the CSS of the `container-item` into a `flex` container. If you want to avoid doing this, let me know, and I'll replace it with a <a href> below the table item, or move the action into the SourceEditView as a button.

(This was actually the main thing why I wanted to change the code here in the first place, talking about being distracted 🤣 )


meeb:
Seems like a nice addition to me, thanks! The FontAwesome icon was missing just because the SCSS is from 2-odd years ago and that icon was added since.

Issue #330, opened 2023-02-12
"Skipping" unavailable media causes it to get unskipped?


kuhnchris:
Hi,
maybe I'm still missing "something" here, but I have following use-case:
One of these aaaaaaaaaaaaages old videos (Mighty Mushin' Emo Rangers) I tried to add/download was (errorously) marked as "not able to be downloaded", so I actually wanted to "skip" the media for now, but it keeps getting re-un-skipped by `signals.py->media_post_save`. 

```
[12/Feb/2023 12:31:51] "POST /media-skip/680dd3de-3e6b-4e40-a82c-6bead3919b7c HTTP/1.1" 302 0
[12/Feb/2023 12:31:51] "GET /media/680dd3de-3e6b-4e40-a82c-6bead3919b7c?message=skipped HTTP/1.1" 200 9613
[12/Feb/2023 12:32:06] "GET /media-skip/680dd3de-3e6b-4e40-a82c-6bead3919b7c HTTP/1.1" 200 6676
2023-02-12 12:32:07,569 [tubesync/INFO] Media: daaaaaaada / _MyJRXfvJmA has a valid publishing date, marking to be unskipped
2023-02-12 12:32:07,574 [tubesync/INFO] Scheduling task to download metadata for: https://www.youtube.com/watch?v=_MyJRXfvJmA
```

![image](https://user-images.githubusercontent.com/761911/218316867-f86f077d-109e-4f2a-964b-fc9fbd2b2bfa.png)

![image](https://user-images.githubusercontent.com/761911/218316891-355eeae2-44f4-4a46-884f-0c11c72a8983.png)

Am I missing something here?

meeb:
You are not missing anything. There's an issue with skipping as mentioned in #280 and other issues. When it was first built the `skip` flag was purely for things like the number of retries and been exceeded and then the media was marked to be permanently skipped internally. This was extended with a UI button that marked media to skip and unskip, however, this is now *also* managed by various other internal conditions where the `skip` flag can be reset if you reset some tasks or update a source and some other events (because if you change the source requirements your skipped media may now be able to be downloaded, so it makes sense to unskip the media so it gets retried again in case it is now available).

Really what this needs is a `skip` flag which is managed internally by tubesync and the downloader, and another `manual_skip` flag which can be set through the web UI so a single `skip` flag isn't re-used for both purposes, and both of these states need to be checked when tasks and signals are run.

kuhnchris:
[...]

PR #334, opened 2023-02-13
Adding new "manual_skip" field; adapt UI

kuhnchris:
solves #330 - please check @meeb 

- Adds a new field "manual_skip" to the Media model.
- Moved and added checks for `manual_skip` (frontend and tasks)
- Cancel/Finish tasks if the media is `manual_skip` (download_metadata)
- Migration(s) for new field
image
(the skip messages differ between 'automatic' and 'manual' skip)
image
image


meeb:
Thanks, a nice start. I think the whole `skip` logic needs a bit of a top-down review, it's a mess as it was hacked in later in development. I'll merge this for now but I have a feeling this may introduce some weird edge cases.

kuhnchris:
Definitly, right now it's a bit of a mess with `skip` and `manual_skip`, I tried to separate the "automation" `skip` from the "manual" `skip`, but we'd probably need to step back a moment and look at the whole thing for the big picture. For now it should be 'fine', if there are edge-cases, I'll try to fix them as soon as I see them popping up on Issues, you can also tag me on them if you want.

Issue #333, opened 2023-02-13
Not existing function `delete_index_source_task`


kuhnchris:
For some reason, `index_source_task`  and `check_source_directory_exists` (`signal.py`) both reference a function `delete_index_source_task` which is nowhere to be found in the project - did you copy that from somewhere and forgot to put it in, or was that on the to-do list and never came around? 
![image](https://user-images.githubusercontent.com/761911/218382430-ba5f7797-7071-4c0c-a55c-de7fe940ad80.png)

(Technically this edge case shouldn't exist, but... you know..)

meeb:
Ah, that did exist at one point, it doesn't now. It's such an edge case that it's extremely rare that it would get called so I guess it was unnoticed until now. These calls can just be removed.

PR #335, opened 2023-02-13
del `delete_index_source_task` calls - solves #333

kuhnchris:
See above. As discussed in #333 it's a dead function left over after some refactoring of the code.


meeb:
Thanks!

PR #337, opened 2023-02-14
align (i) better with text (+ checkbox less wide)

kuhnchris:
Make the (i) less stuck to the text.

before:
image

after:
image

(yes, this is another feature I work on in the 1st screenshot, sorry, I only had that screenshot on hand right now.)
(also, yes, this is cosmetic, so low priority) 🔽

### DIFF
--- a/tubesync/common/static/styles/materializecss/components/forms/_checkboxes.scss
+++ b/tubesync/common/static/styles/materializecss/components/forms/_checkboxes.scss
@@ -14,7 +14,7 @@
   // Text Label Style
   + span:not(.lever) {
     position: relative;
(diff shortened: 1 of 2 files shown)


meeb:
Fine with me, the only thing with this is that the changes to `_checkboxes.scss` would be overwritten if materializecss was ever updated.

kuhnchris:
Well, I could make a scss overwrite in the `tubesync.scss`, would you prefer that?

meeb:
Probably might be a bit cleaner, yes. Although I don't believe materializecss is even being updated any more so it's probably not that important.

PR #374, opened 2023-05-02
Fix #364 by checking the filesize is not null.

### LINKED ISSUE #364: Media import with .mp4 files causes odd issues and Latest Downloads to be full of garbage
I ran a media import of my existing youtube channels when setting up tubesync, and it correctly found my .mp4 files, and my .mkv files.  However, now, I have 2 strange issues:

1) My latest downloads list on the dashboard, is full of ancient videos that were imported, with no dates.

2) If I go look at one of those videos, it seems to think the file is a .mkv, when it's actually a .mp4, and then tells me the file size is 0 (in media details).

My guess is that it found the video with the .mp4 name, imported it, and then assumed it had a .mkv name when making the database entry?

I guess I don't care that the media info pages don't work, I just want the latest downloads to work, maybe ignore entries that the system thinks are 0-bytes?

garbled1:
This fixes the latest downloads page to continue to work even after you have manually imported media.

### DIFF
--- a/tubesync/sync/views.py
+++ b/tubesync/sync/views.py
@@ -70,7 +70,7 @@ class DashboardView(TemplateView):
             data['average_bytes_per_media'] = 0
         # Latest downloads
         data['latest_downloads'] = Media.objects.filter(
-            downloaded=True
+            downloaded=True, downloaded_filesize__isnull=False
         ).order_by('-download_date')[:10]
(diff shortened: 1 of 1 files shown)


garbled1:
Fixes #364

meeb:
Thanks!

PR #381, opened 2023-05-26
in cleanup_old_media, filter in database rather than in Python

a-kr:
Current implementation of cleanup_old_media loads *all* downloaded media objects into memory and then filters them in code. This results in large memory consumption (5+ GB for a library of 15K+ media items).

This PR moves filtering to the database query.


meeb:
Thanks! Looks fine to me, I'll merge it into the next release. A lot of the clean-up code like this dates when tubesync was my personal hacky download client so efficiency PRs are welcome.

Issue #367, opened 2023-04-07
[Feature Request] Add subtitle download support 


mylogon341:
It would be great if there was another checkbox in the sources screen that asked if you want to download subtitles too. Not only would this be beneficial for accessibility, but, for example, the channel [Primitive Technology](https://www.youtube.com/@primitivetechnology9550) put's all his "narration" in the subtitles, so it makes channels like that usable.

As for the technical side, yt-dlp supports this quite well with its [subtitle docs](https://github.com/yt-dlp/yt-dlp#subtitle-options) mentioning options.

Thanks a lot

meeb:
This is already an open issue. It's moderately complicated to add subtitles as a feature, it's not just as simple as passing in `--write-subtitles` as tubesync will need to keep track of the downloaded files, index available languages and support the post-processing filters for optional burn-in. See #7 which is the main issue to track this. It's only been recently sufficient features are available upstream to easily add and manage subtitles. I'll close the issue for now as a duplicate.

meeb:
Also, thank you for the support, most appreciated!

mylogon341:
Ah. Sorry. I was on sorting out a few things yesterday and must have thought I had searched for a duplicate of this but clearly didn't. 
Welcome for tip though - you're doing good work here and deserve that + more attention as you continue to improve.

PR #406, opened 2023-09-02
Subtitles

pacoccino:
I propose this code changes to enable subtitles downloading. 
It is the first time for me using python&django so it’s a bit of discovery ;)

I don’t know if I’ve done everything that would have been necessary to add this feature, but it does what I need as it is. 

I worked on this as I really need to have subtitles downloaded and I don’t need/want something as heavy as tubearchivist. 

#7 #367


meeb:
Thanks for the contribution! This looks fine to me. The only thing I would suggest is some model form validation as a Django validator that confirms that the subtitle language field is actually a comma-separated array of two letter, lower case values. Currently as far as I can see this will otherwise let people pass arbitrary data to `yt-dlp`'s API which is probably worth putting some guide rails on.

Something like (untested, just typed ad-hoc):

```python
def validate_is_subtitle_lang_list(value):
    parts = value.split(',')
    if not parts:
        return True
    subtitle_lang_chars = 'abcdefghijklmnopqrstuvwxyz'
    for i, lang in enumerate(parts):
        if len(lang) != 2:
            raise ValidationError(f'Subtitle lang code at position {i} is {len(lang)} chars, must be 2 chars')
        for c in lang:
            if c not in subtitle_lang_chars:
                raise ValidationError(f'Subtitle lang code at position {i} must be a-z lowercase only, got {lang}')
    return True
``` 

As you're new to Django, here's how validators are integrated to save you looking it up: https://docs.djangoproject.com/en/4.2/ref/validators/ - you can add the validator to your new `sub_langs` field. Thanks again for the PR, most appreciated.

pacoccino:
Nice catch, I added a validator, but opted for a regex instead, matches better the [options available](https://github.com/yt-dlp/yt-dlp#subtitle-options)

Another thing is that it won't download subtitle for already downloaded videos, I doubt you can use yt-dlp just for downloading subtitle without video but I'll check that

meeb:
Nice, thanks! I'll give this a test and merge it into the next release.

tehniemer:
Will this grab subtitles for already downloaded media?

meeb:
Alas, no it won't.

Issue #400, opened 2023-08-17
Feature Request: Delete videos that have been removed from a list


frob:
It would be nice if I could delete the video from my playlist and have it be removed from my media library as well.

This would be an alternative to the Download Cap and the Delete after Days feature.

tcely:
This currently exists with two options: 1) delete the video from tubesync and/or 2) delete the media file from the disk.

I think this issue can be closed now.

meeb:
Yep seems resolved.

PR #420, opened 2023-10-20
Adds workflow to delete local media that no longer exists in the source

sparklesmcfadden:
This adds the option requested in feature request #400, removing media that has been removed from a playlist.


meeb:
Looks good to me, thanks!

PR #422, opened 2023-10-20
Added a regex filter string to sources and performs a regex skip if t…

locke4:
…he filter string is not matched in the video title


locke4:
Trying to work out why this is merging into main branch and not my fork... my bad

PR #423, opened 2023-10-20

ltomes:
While migrating to an external database to get around database locked errors I needed this minor change to load in my database.


meeb:
Thanks!

Issue #424, opened 2023-10-21
Skip videos whose titles don't match a regex string pattern


locke4:
This is a feature request for the skip flag to be set if a video title does not match a user-specified regex pattern.

I want to add a channel whose videos titles are prefixed with the topic. I'm only interested in downloading those specific videos. This feature would allow me to filter the downloaded videos. I'm working on a fork to add this support which seems to be working but I need to write tests for. Once it has a suitable test in tests.py, I'll submit a PR for review by a maintainer and reference this feature request.

meeb:
Sounds like a reasonable addition, happy to review the PR once you're ready.

perry-mitchell:
@meeb @locke4 Can this be closed? I can see it in the UI now.

meeb:
Yep!

Issue #183, opened 2021-11-10
Old videos don't download


freddy2659:
Using the "Get next best resolution or codec" option doesn't retrieve videos that are very old (2009 at 240p best resolution).

Example channel ID: UCDsElQQt_gCZ9LgnW-7v-cQ
Example video: https://www.youtube.com/watch?v=QCUCTLk05cA

The reason provided: Media cannot be downloaded because it has no formats which match the source requirements. 
Available formats: Media has no indexed available formats
Desired format: 1080p (video:VP9, audio:OPUS) 60FPS HDR

I'm not sure if one of the flags (60FPS, HDR, OPUS, VP9. 1080p) is causing it to ignore the video that IS available? Or if it's just because the video doesn't provide the information expected.

It could also be because it is both the next best resolution AND codec?

meeb:
Thanks for the issue. Yeah that should be matched, it indeed has a 240p version and the cut-off was lowered to 240p in #162 so something has changed in the metadata most likely that's causing the matching code to require a tweak.

azukaar:
Noticed the same issue, the log does have additional infos tho: "has no published date set, marking to be skipped" on old videos, may be some older videos have a different format for their metadata that TubeSync does not understand

meeb:
@azukaar can you paste a link to a video that has that error? If it is legitimately missing a publish date in the metadata that would cause this issue.

azukaar:
Hello! weirdly enough, I was going back to Tubesync to get such videos, all the videos marked as skipped by the system seems to have been downloaded anyway now, despite of having that message in the log "has no published date set, marking to be skipped",
Regardless here's an example of a video that caused this log, and was marked as skipped, but that eventually got downloaded anyway: /watch?v=NNhzYkXE1i4
Note that unlike what I initially thought, this is a fairly recent video so the issue might not be related, and may be is just a logging issue.

```
2022-02-02 19:27:25,956 [tubesync/WARNING] Media: Todo / NNhzYkXE1i4 has no published date set, marking to be skipped
2022-02-02 19:27:26,120 [tubesync/INFO] Scheduling task to download metadata for: https://www.youtube.com/watch?v=NNhzYkXE1i4
2022-02-02 19:27:26,308 [tubesync/INFO] Indexed media: Todo / NNhzYkXE1i4
```

meeb:
[...]

PR #425, opened 2023-10-21
Add support for regex video title filtering

locke4:
Feature addition to add filtering of video titles by regex, implementing #424 

First time working with Django so please let me know any improvements that can be made! I looked into field validation and null handling but struggled to find a cleaner way to validate form input data. I also couldn't find a way to get the filter working for the test youtube channels in tests.py, so currently all tests run with the regex string "*.". I've tested it manually with a couple of channels and the regex string seems to work correctly.

I also needed to rework the "if not downloaded yet" behaviour in signals.py to ensure the behaviour of editing a source prior to an actual download occurring also supports this regex. I've done some testing and it *seems* like the behaviour is as expected but this may need more testing to ensure I haven't introduced any bugs by accident.


locke4:
Also fixes a bug with the media.html pagination and adds support for triggering workflows from dispatch triggers.

meeb:
Thanks for the contribution! Mostly looks good, there's a couple of things that would probably need to be tweaked before it gets merged. I'll comment in-line on the commits.

tubesync/sync/migrations/0001_initial.py
```diff
@@ -24,6 +24,7 @@ class Migration(migrations.Migration):
                 ('source_type', models.CharField(choices=[('c', 'YouTube channel'), ('p', 'YouTube playlist')], db_index=True, default='c', help_text='Source type', max_length=1, verbose_name='source type')),
                 ('key', models.CharField(db_index=True, help_text='Source key, such as exact YouTube channel name or playlist ID', max_length=100, unique=True, verbose_name='key')),
                 ('name', models.CharField(db_index=True, help_text='Friendly name for the source, used locally in TubeSync only', max_length=100, unique=True, verbose_name='name')),
+                ('filter_text', models.CharField(db_index=True, help_text='Regex compatible filter string for video titles', max_length=100, verbose_name='filter text')),
```
meeb: Did you manually add this to the `0001_initial.py` file? Migrations need to be created by `make_migrations` generally. Adding this here will break everyones installations when `migrate` is run on their existing databases.

.github/workflows/ci.yaml
```diff
@@ -4,6 +4,7 @@ env:
   IMAGE_NAME: tubesync
 
 on:
+  workflow_dispatch:
```
meeb, line 7: What's the purpose of this?

tubesync/sync/models.py
```diff
@@ -287,6 +287,13 @@ class IndexSchedule(models.IntegerChoices):
         help_text=_('If "delete old media" is ticked, the number of days after which '
                     'to automatically delete media')
     )
+    filter_text = models.CharField(
+        _('filter string'),
+        max_length=100,
+        default='.*',
```
meeb: This should probably default to blank and not run a regex at all unless specifically set, otherwise this will be running a `.*` pointless regex against every item.

tubesync/sync/signals.py
```diff
@@ -104,36 +105,57 @@ def media_post_save(sender, instance, created, **kwargs):
...
         max_cap_age = instance.source.download_cap_date
-        published = instance.published
-        if not published:
-            if not instance.skip:
-                log.warn(f'Media: {instance.source} / {instance} has no published date '
-                         f'set, marking to be skipped')
+        filter_text = instance.source.filter_text
+        published = instance.published 
+
+        if instance.skip:
```
meeb, line 110: Can you fix the whitespace here? Rest of the project uses 4 spaces.

tubesync/sync/signals.py
```diff
@@ -104,36 +105,57 @@ def media_post_save(sender, instance, created, **kwargs):
...
+                    elif max_cap_age:
+                        if published > max_cap_age:
+                            # Media was published after the cap date but is set to be skipped
+                            log.info(f'Media: {instance.source} / {instance} has a valid '
+                                    f'publishing date, marking to be unskipped')
+                            instance.skip = False
+                            cap_changed = True
+                    elif filter_text:
+                        if re.search(filter_text,instance.title):
+                            # Media was published after the cap date but is set to be skipped
```
meeb: Comment needs an update.

tubesync/sync/tasks.py
```diff
@@ -254,6 +255,11 @@ def download_media_metadata(media_id):
             log.warn(f'Media: {source} / {media} is older than cap age '
                      f'{max_cap_age}, skipping')
             media.skip = True
+    # If the source has a search filter, check the video title matches the filter
+    if not re.search(source.filter_text,media.title):
```
meeb: Should this be an opt-in filter (only download media that match this regex) or an opt-out filter (skip media that match this regex)?

tubesync/sync/tasks.py
```diff
@@ -254,6 +255,11 @@ def download_media_metadata(media_id):
...
+        # Filter text not found in the media title. Accepts regex string, blank search filter results in this returning false
+        log.warn(f'Media: {source} / {media} does not contain {source.filter_text}, skipping')
```
meeb: Nitpick, but this is a match not a contain.

tubesync/sync/tests.py
meeb, line 1: This PR does need a functional test suite method. Perhaps you could add a `Source.is_regex_match(media_item_title)` style method, use that everywhere in `tasks.py` and `signals.py` rather than `re` directly and write a `test_is_regex_match` method to confirm it's working with 5-10 basic regex examples?

.github/workflows/ci.yaml
locke4: This is to allow me to trigger the workflow from Git CLI with `gh workflow run` or from the web UI. It was  so I could manually run the workflow on my feature branch before merging it into my fork main branch. Is there an easy way to reject just this change for the PR as it's not necessary?

image

tubesync/sync/migrations/0001_initial.py
locke4: Yes I did add this here in my editor, I'm not familiar with this framework but I can see why this would be an issue! Could you point me to info on how to fix this?

tubesync/sync/models.py
locke4: I struggled a bit with default blank behaviour and this was my fix, but I can see why this is undesirable. I'll put an "if filter_text:" to suppress wherever the regex is called for the blank behaviour and double check anywhere this doesn't work.

tubesync/sync/tests.py
locke4: Great idea, I'll look into this

tubesync/sync/signals.py
locke4: Updated

tubesync/sync/tasks.py
locke4: I thought opt-in was more intuitive for a user. Currently the filter_text is specified, and then if anything does not match the filter, it is marked as a "skip". Happy to take your suggestion to this though as once I re-write this to use an is_regex_match(media_item_title) method it'll be easy to change later. Certainly regex supports inverse matching though it's clunky. The other way would be to specify both an "include" and an "exclude" string. I.e. include "foo" but exclude "bar" but it feels unnecessarily complex. On my local fork I'm already using a regex to download all videos which contain "40k" but don't contain "Darktide".

```
foo -> match
bar -> not matched
foo bar -> not matched
```

tubesync/sync/tests.py
locke4: Written the test suite but struggling to get the expected results to match my actual behaviour. For some reason the filter_text 'No fancy stuff' matches with 'no fancy stuff title'... Have to look at this again later in the week as I can't work it out. Quick local script to test works fine so assuming I'm missing something obvious here.


tests.py:
```
    def test_is_regex_match(self):
        self.media.metadata = all_test_metadata['boring']
        expected_matches = {
            ('.*'): (True),
            ('no fancy stuff'): (True),
            ('No fancy stuff'): (False), #fails here? re.search('No fancy stuff', 'no fancy stuff' is returning true when by default, re.search should be case sensitive
            ('(?i)No fancy stuff'): (True), #set case insensitive flag
            ('no'): (True),
            ('Foo'): (False),
            ('(?=.*f.*)(?=.{0,2}|.{4,})'): (True),
            ('f{4,}'): (False),
            ('^[^A-Z]*$'): (True),
            ('^[^a-z]*$'): (False),
            ('^[^\\s]*$'): (False)
        }

        for params, expected in expected_matches.items():
            filter_text = params
            expected_match_result = expected
            self.assertEqual(self.source.is_regex_match(self.media.title), expected_match_result)
```
models.py:
```
    def is_regex_match(self, media_item_title):
        if re.search(self.filter_text,media_item_title):
            return True
        else:
            return False   

```

locke4: Using following test vector with metadata.json, currently passing
```
expected_matches = {
            ('.*'): (True),
            ('no fancy stuff'): (True),
            ('No fancy stuff'): (False),
            ('(?i)No fancy stuff'): (True), #set case insensitive flag
            ('no'): (True),
            ('Foo'): (False),
            ('(?=.*f.*)(?=.{0,2}|.{4,})'): (True),
            ('f{4,}'): (False),
            ('^[^A-Z]*$'): (True),
            ('^[^a-z]*$'): (False),
            ('^[^\\s]*$'): (False)
```

tubesync/sync/migrations/0001_initial.py
meeb: With Django you just make the adjustments in `models.py` and then run `./manage.py make_migrations`. The `migrations/XXXX_name-here.py` files are created automatically. Once you have a migration you run `./manage.py migrate` and the migrations are checked and applied one at a time in order. You can edit them manually if you need to, but they are in a specific order so if you edit old ones they break the migration chain.

Docs here:

https://docs.djangoproject.com/en/4.2/topics/migrations/

You would need to remove this edit, then use `make_migrations` to create a new migration for this PR.

.github/workflows/ci.yaml
meeb: Fair enough, I just generally use `make test` locally. The CI tests were to confirm it was working before I build a release image. Providing this doesn't break anything it's fine to leave it.

Honestly, I don't use GitHub actions much so this wasn't a critique I didn't actually know what this did.

tubesync/sync/tasks.py
meeb: I would imagine the most common use-case for this is going to be "I want to ignore specific videos on a channel". While you can do this with positive matching regexes having to use lots of `(?!string$)`'s everywhere isn't going to be that friendly.

Personally, I think this might be better as a `exclude_regex` field. What do you think?

locke4: I'm not sure I follow. Negative lookaheads get this behaviour you describe. Inverting it is no work though I want to check that re supports negative lookaheads as I might need to swap it for re2. My primary use case is to do both (include the word "foo" but not include the word "bar") as above.

```
Regex string: ^(?!.*foo).*$

foo
bar <<< Match
foo bar

Regex string: foo

foo <<< Match
bar
foo bar <<< Match
```

tubesync/sync/migrations/0001_initial.py
locke4: All done, I noticed when I ran makemigrations that it altered two fields from a PR a month ago which I didn't touch in my repo. Hoping that's not an issue as doesn't seem to be from this PR

auto_subtitles:
model.py has a hyphen not in the migration file for that PR

sub_langs:
model.py has a regex validator not in the migration file for that PR

locke4:
Merged branch addressing all comments into this PR

tubesync/sync/tasks.py
perry-mitchell: Why not add an option to specify what the filter does? Include/Exclude.

For various use cases having an allow-list vs an ignore-list might be very powerful.

Imagine a channel with thousands of videos, and perhaps a reoccuring _title_ that appears throughout the channel, like "office update 15". It would be far easier to just set an include list of `^office update \d+$`, imo.

Or perhaps a channel that is otherwise about some topic, but has music videos strewn about it.. You might set an exclude list of `\(music\svideo\)`.

Writing regex is hard for some users and negations are tricky to reason about. Avoiding the _need_ for them would be advantageous imo. Requiring one extra switch isn't a lot of work.

locke4: I like this idea. Essentially just a dropdown or slider toggle "include/exclude" though so it's clear what behaviour is currently selected (tickboxes would be a bit ambiguous. I'll have a look later today or tomorrow at do a new branch to implement this.

meeb:
Thanks for your work here and sorry for the delay, I was away for a while. I'll review this and merge it into the next release.

locke4:
> Thanks for your work here and sorry for the delay, I was away for a while. I'll review this and merge it into the next release.

Thanks and no worries, I've also been away! I was working on a new branch to implement the toggle between include/exclude as suggested above, but it had a few bugs I couldn't track down so it hasn't been merged yet. Once I've tracked that down I'll submit a new PR to get feedback.

meeb:
Thanks! Submit a PR for that when you're ready, it'll probably be easier to review all in one.

meeb:
This PR introduced logic bug in the handling of skipping media (which is at best a mess right now so understandable) as well as a race condition where metadata was checked for the upload date before the metadata had been downloaded, marking every media item as initially skipped. I've pushed a fix here:

https://github.com/meeb/tubesync/commit/e54a762a7bd93d8070a2b1057f16c131cb509755

You don't need to do anything and this seems to be working again, just mentioning for reference.

Was reported in #440 and #183 among other comments.

perry-mitchell:
How is this meant to work, in terms of filtering _out_ videos? I suppose my regex is still going to match everything in this channel I'm processing with this:
image

Issue #206, opened 2022-01-19
Not deleting media older than "days to keep"


rrediske:
First, thank you for this wonderful way to download videos automatically!

The problem I am having is that tubesync isn't removing videos from any of the four channels I have it watching, so I will eventually run out of disk space if I can't find a way to delete old videos. I used your basic 13 line docker-compose.yml, so I have nothing special for configuration. When adding each channel, I set the "download cap" to 1 week and the "days to keep" to 10 days, but tubesync still has every video it's ever downloaded going back to December 13th of last year (I'm now at 63 videos).

In a previous install on another machine, I tried deleting videos manually from the mounted volume "tubesync-downloads", but that seemed to cause tubesync to stop downloading anything else, so I wound up moving to a new machine to start over. It takes almost 2 full days for tubesync to index all the videos of the 4 channels I watch, so I really want to avoid having to do that.

Any ideas? Am I doing something wrong?

meeb:
Thanks for the comments! That could well be a bug. I'll leave this open and investigate. I'm assuming you're on normal-ish Linux and not running it with weird WSL paths on Windows or anything?

rrediske:
Open SuSE Leap 15.3, so... weird enough, but normal-ish :) I'm running a home assistant and a nextcloud docker image in the same machine and they have been fine for a few months. It's a VM on a Dell R710, running ESXi, 8 GB RAM, 120 GB disk.

meeb:
Thanks for the details. If you can easily search the container logs are there any errors? If for some reason it's attempting to clear up files that don't exist due to an invalid path or similar issue there should be a log of it.

rrediske:
I did ```docker logs 8703 >& abc``` then did a grep ERROR on that and I got 39 errors in the last 24 hours of the form:
```
2022-01-19 19:46:49,576 [tubesync/ERROR] ERROR: [youtube] 7wLxM7oNN1s: This video is unavailable on this device.
```

39 sounds like the number of videos that should be getting deleted, might be a coincidental number, though.

Here's more context:

```
Rescheduling task Downloading metadata for "21e4dd9c-6e1e-4bff-ac85-df44d526a059" for 5:45:41 later at 2022-01-19 11:35:05.305812+00:00
2022-01-18 23:49:28,602 [tubesync/DEBUG] [youtube] 7wLxM7oNN1s: Downloading webpage
[...]

PR #426, opened 2023-10-22
Delete files on disk

sparklesmcfadden:
Adds option to delete files on disk when removing a media item from TubeSync. Restores previously-removed code in signals.py (as mentioned in #206) and gates it behind an option.


meeb:
Thanks for the contribution! This looks reasonable to me. There also may be subtitles and metadata files stored on disk as well (`vtt` and `nfo` files etc.). If you could update this to also clean those up when media is deleted I'll merge this into the next release.

hunterpankey:
@meeb or @sparklesmcfadden Will we need to manually delete media/files for items that have already been removed from TubeSync, i.e., orphaned files on disk? I just updated and forced a reindex for a few channels and files aren't going away.

meeb:
Yes, if you've deleted the record of the media item in tubesync it won't be cleaned up on disk automatically. tubesync doesn't scan existing files for orphans and delete them.

PR #435, opened 2023-11-19

klinker41:
Add documentation about how to use a docker compose postgres container and connect it to tubesync. This seems like a fairly basic use case that many users would want to implement, given the large performance benefits it brings.


meeb:
Thanks! Looks like a good addition to the docs.

PR #443, opened 2023-12-04
compress sync_media table for mariadb

depuhitv:
Ran into large database table sizes after importing a large channel and noticed it was the sync_source which is storing the entire metadata file content for some reason.

Not sure if thats used later at some point so just modified the database to compress the column instead.
Tested and works with Mariadb but assume the same would work for mysql as well.

### DIFF
--- a/docs/other-database-backends.md
+++ b/docs/other-database-backends.md
@@ -81,7 +81,8 @@ you are now using an external database server for your TubeSync data!
 
 ## Database Compression (For MariaDB)
 With a lot of media files the `sync_media` table grows in size quickly.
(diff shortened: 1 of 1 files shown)


meeb:
Thanks for the PR! The full metadata is kept because when you change source parameteres (e.g. desired resolution) you need to re-evaluate the metadata to select which other available streams are most suitable. Same with the subtitles and audio language data. You could probably shave off some of the metadata before storing it (such as maybe some of the thumbnail formats) but as a lot of it is re-used regularly I just stored all of it.

Happy to include your notes for the docs though, thanks.

Issue #379, opened 2023-05-15
TubeSync metada to Jellyfin


marciglesias17:
I am trying to get TubeSync to download full playlists only the audio and Jellyfin reads the NFO, but it is not doing it, it seems the NFO format is not the same, any idea how I can do this?

marciglesias17:
Is there any way to make it compatible with Jellyfin's NFO?

meeb:
Manually? Sure just edit the NFO in a text editor. For automatic integration it looks like I need to tweak the NFO format.

marciglesias17:
Ok, I will wait for your update, do you plan to do it soon?

meeb:
Honestly no idea, I need to review what needs to be done properly first. If it's a quick patch with just some different file names, probably. If it's more work it'll go onto the wishlist.

marciglesias17:
Okay, keep me informed. Thanks!

meeb:
Happy to accept pull requests if you want something in the project.

breakid:
As an additional datapoint, I tested tubesync video NFO files with Emby, a sibling project to Jellyfin, and they worked fine. I compared the files, and, besides the element order, the only differences I saw were the Emby NFO had:

- `standalone="yes"` in the doctype tag (i.e., `<?xml version="1.0" encoding="utf-8" standalone="yes"?>`)
- The contents of the `<plot>` tag wrapped with a CDATA section (i.e., `<plot><![CDATA[The plot goes here]]></plot>`

meeb:
I don't think either of those differences would affect the NFO file from being read unless Jellyfin is extremely picky. I'll add `standalone="yes"` to the doctype, it can't hurt. I don't think that'll fix this for you though.

InterN0te:
Hi !
I just set up Tube Sync to try it with Jellyfin and it's f*cking greats !
The only change I could suggest would be to add the Season and Episode tags in the NFO:
```
   <showtitle>Video Name</showtitle>
   <season>{yyyy}</season>
   <episode>{X}</episode>
``` 

With {yyyy} being the year of publication and {X} being 1 for the first video of that year and then counting in order of publication

InterN0te:
With this small modification in the NFO and Jellyfin which scans the download directory (the library is a "TV Series"), Jellyfin correctly displays the channels and is able to sort videos :
![image](https://github.com/meeb/tubesync/assets/10882916/03f8e2b0-dd0e-4c93-b878-14b1468be2a5)
[...]

PR #447, opened 2023-12-09
Added season and episode tags in NFO to get Jellyfin displaying correctly

InterN0te:
Following #379 :
For Jellyfin to get information to sort the videos, I added the season and episode tags in the NFO file:

The season is always 1 if the source is a playlist and is the year of publication if the source is a channel

The episode is the position in the playlist or the number of the video in the year, sorted by publication date if the source is a channel

I tested it with a Docker container and found that it worked correctly but I couldn't run the test script so I changed the expected NFO without having it verified


InterN0te:
Redone in #448 with branch renaming

PR #448, opened 2023-12-09
Added season and episode tags in NFO to get Jellyfin displaying correctly

InterN0te:
Following #379 :
For Jellyfin to get information to sort the videos, I added the season and episode tags in the NFO file:

The season is always 1 if the source is a playlist and is the year of publication if the source is a channel

The episode is the position in the playlist or the number of the video in the year, sorted by publication date if the source is a channel

I tested it with a Docker container and found that it worked correctly but I couldn't run the test script so I changed the expected NFO without having it verified

### DIFF
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -1269,6 +1269,21 @@ class Media(models.Model):
         showtitle.text = str(self.source.name).strip()
         showtitle.tail = '\n  '
         nfo.append(showtitle)
+        # season = upload date year
+        season = nfo.makeelement('season', {})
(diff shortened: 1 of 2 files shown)


InterN0te:
For a Youtube Channel :
image

For a playlist :
image

The only thing that could be added would be the channel thumbnail as a thumbnail of the TV show and seasons with 2 files :
- poster.jpg (TV show thumbnail)
- season-poster.jpg (thumbnail for all seasons)

When done manually on the TV Channel, here is the result (I took a video thumbnail to do it) :
image

meeb:
This looks good to me. I don't really have any additional comments, the format and where you've edited are fine. Manually adjusting the test makes sense here as well. I'll merge this into the next release. Thanks for the contribution!

PR #450, opened 2023-12-11
Following Delete files on disk #426

InterN0te:
I couldn't figure out how to edit the code in PR #426 (I think I'm not allowed to) so here I am !

I took its fork and updated it from the main branch to rework the file deletion. It now removes:
- for media, find all files matching the filename without extension when source option ‘Delete file from disk’ is enabled or if removed with 'Delete and skip media' button
- for sources, all media (with previous method) then the folder and all its remaining content if ‘Also delete downloaded media’ is checked


meeb:
Thanks. Looks neater. I'm not sure these needs an additional source-level checkbox?  There is already (limited) functionality for this that could be replaced or extended when you delete a source (the "also delete downloaded media" checkbox):

https://github.com/meeb/tubesync/blob/main/tubesync/sync/views.py#L420

Given there currently isn't a manual option to delete media at the moment would this likely only be triggered when the source is deleted? I don't object to the deletion being moved to the signal, that's probably where it should originally have been.

What's your original full use case for this feature? If you can explain when and how you delete media we can likely merge the features to be more sensible.

I'm also a little concerned about the number of new flags and checkboxes at the source level getting out of hand after the last few PRs!

InterN0te:
I was also skeptical about this option but it is actually triggered by the settings (already present it seems to me) "Delete old media" and "Days to keep" or by "Delete removed media" so I kept it
With this, the user can choose to clean TubeSync but keep the videos for their media server or keep the folder sync with TubeSync
Actually, I don't really know if it's useful to keep deleted TubeSync videos in the folder, but I didn't want to force such a change


I'm not used to developing in python and I don't know Django, if the code requires reorganization, don't hesitate

My use case would be to automatically download videos from a few channels to find them in Jellyfin which will track it for me, tell me when a new one is available, ... and give me about 30 days to watch the videos before they are automatically deleted from TubeSync and the disk.
So I think I need to use "Delete Old Media" and "Days to Keep Media", but so far it doesn't seem to delete them from disk and doesn't want to force everyone

meeb:
Thanks, your use case makes sense. I'll give the code a more thorough review.

meeb:
This hasn't had any issues I've noticed in testing so I'll merge it into the next release. Thanks for your contribution!

meeb:
Yep probably!

PR #452, opened 2024-01-01

ShaneBridges1234:
Corrected table name in SQL for MariaDB column compression.


meeb:
Nice catch, thanks!

PR #460, opened 2024-01-16
Optimized source page and dashboard loading

skayred:
I have faced the issue with several pages - sources view and the dashboard - when I have relatively big channels synced. I have checked the queries and found out, that the column `metadata` is huge and makes the query extremely slow in the places where it is not required (100 ms vs 2-7 seconds). The easiest workaround is to exclude the field from the selection.


meeb:
Excellent, thanks! Looks sensible to me.

Issue #468, opened 2024-02-21
Processing when a downloaded video is changed to private


woozu-shin:
### Version
- 0.13.3

### Abstract
- Processing when a downloaded video is changed to private, will raise exception and hang whole background tasks.
- Add exception handling when retrieving media info if the original video is not accessible.

### Details
- traceback log
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/background_task/tasks.py", line 43, in bg_runner
    func(*args, **kwargs)
  File "/app/sync/tasks.py", line 238, in download_media_metadata
    metadata = media.index_metadata()
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/sync/models.py", line 1422, in index_metadata
    return indexer(self.url)
           ^^^^^^^^^^^^^^^^^
  File "/app/sync/youtube.py", line 60, in get_media_info
    raise YouTubeError(f'Failed to extract_info for "{url}": No metadata was '
sync.youtube.YouTubeError: Failed to extract_info for "https://www.youtube.com/watch?v=REDACTED": No metadata was returned by youtube-dl, check for error messages in the logs above. This task will be retried later with an exponential backoff.
```

### References
- n/a

woozu-shin:
Is there any branches git-flow related?
In my opinion, it is a medium-sized project, and one that is receiving interest from many users.
If there is branch strategy, more active P/R may be possible. :)
thx for nice project.

meeb:
Exceptions raised by the worker won't block tasks and they are designed to be informative traces in the container logs so raising exceptions on failures is working properly as designed. I'll close this for now.

PR #469, opened 2024-02-21
[ISSUE-468] Fix infinite error when accessing an invalid video which is already downloaded

woozu-shin:
### Issue
- #468 

### Descrption
- When accessing an invalid video which is already downloaded
  - Mark as can not downloadable, skippable. then save.
- This may occur in the following cases:
  - 1) Add source and sync all media.
  - 2) Youtube creator changed specific video to private.
  - 3) Re-sync media source (like metadata..)
  - 4) Background task is not resolved

### References
- n/a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,6 @@ dmypy.json
 
 Pipfile.lock
 .vscode/launch.json
+
+# Jetbrains
+.idea
--- a/tubesync/sync/signals.py
+++ b/tubesync/sync/signals.py
@@ -172,7 +172,7 @@ def media_post_save(sender, instance, created, **kwargs):
         instance.save()
         post_save.connect(media_post_save, sender=Media)
     # If the media is missing metadata schedule it to be downloaded
-    if not instance.metadata:
+    if not instance.metadata and (instance.can_download and not instance.skip):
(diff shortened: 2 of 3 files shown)


meeb:
Thanks for the PR! This isn't a bad idea and your issue does need to be fixed, however automatically skipping every video that fails to download metadata for any reason is probably a bit over zealous, for example if you use a proxy server that has a wobble or YouTube rate limits your IP then every single queued video will be marked to skip and not retry. Currently I don't believe it's an infinite error, the workers will retry multiple times over a day or two and then mark the media item to skip.

woozu-shin:
> Thanks for the PR! This isn't a bad idea and your issue does need to be fixed, however automatically skipping every video that fails to download metadata for any reason is probably a bit over zealous, for example if you use a proxy server that has a wobble or YouTube rate limits your IP then every single queued video will be marked to skip and not retry. Currently I don't believe it's an infinite error, the workers will retry multiple times over a day or two and then mark the media item to skip.

Oh, I see. I noticed a lot of exceptions.
This is why manually skip exists, and I think you can use that function.

meeb:
I don't think this PR is needed given it's just lots of retries and not an infinite loop. I'll close this for now but feel free to re-open or comment further if needed.

PR #471, opened 2024-02-21
Adding auth option for healthcheck

Re4zOon:
Its grabbing the HTTP_USER and PASSWORD env variables and using them for authentication for the healthcheck.
Currently it fails if the auth is enabled, marking the container as unhealthy.

Tested with both set and unset variables.

### DIFF
--- a/tubesync/healthcheck.py
+++ b/tubesync/healthcheck.py
@@ -12,24 +12,38 @@
 
 import sys
 import requests
-
+import os
 
 TIMEOUT = 5  # Seconds
 
-
-def do_heatlhcheck(url):
+def do_heatlhcheck(url, username, password):
     headers = {'User-Agent': 'healthcheck'}
-    response = requests.get(url, headers=headers, timeout=TIMEOUT)
+    if username and password:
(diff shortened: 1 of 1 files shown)


tubesync/healthcheck.py
```diff
@@ -12,24 +12,38 @@
...
     headers = {'User-Agent': 'healthcheck'}
-    response = requests.get(url, headers=headers, timeout=TIMEOUT)
+    if username and password:
+        response = requests.get(url, headers=headers, timeout=TIMEOUT, auth=(username, password))
+    else:
+        response = requests.get(url, headers=headers, timeout=TIMEOUT)
     return response.status_code == 200
 
+def get_http_auth():
+    if "HTTP_USER" in os.environ:
```
Re4zOon, line 28:
```suggestion
    if "HTTP_USER" in os.environ and "HTTP_PASS" in os.environ:
```

Re4zOon: Adding the PASS as well, just in case someone only sets one or the other.
Healthcheck will be unsuccessful, but at least python will not crash :)

Re4zOon:
Its not the cleanest solution, but it does the job. Its 9 pm after a long day...

PR #472, opened 2024-02-21
Adding HTTP auth

Re4zOon:
Its grabbing the HTTP_USER and PASS env variables and using them for authentication for the healthcheck.
Currently it fails if the auth is enabled, marking the container as unhealthy.

Tested with both set and unset variables.

After I opened https://github.com/meeb/tubesync/pull/471 and got disgusted with it, I asked chatgpt to write one instead...

### DIFF
--- a/tubesync/healthcheck.py
+++ b/tubesync/healthcheck.py
@@ -1,35 +1,32 @@
-#!/usr/bin/env python3
-'''
-
-    Perform an HTTP request to a URL and exit with an exit code of 1 if the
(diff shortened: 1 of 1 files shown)


Re4zOon:
Both of them are still in place :)

https://github.com/Re4zOon/tubesync/edit/patch-2/tubesync/healthcheck.py

meeb:
I'm assuming ChatGPT or something wrote this? I'll just add the feature myself to preserve formatting, features and general sanity.

PR #480, opened 2024-03-03
use env vars for source defaults

rpowel:
Check for env variables for source setting defaults when adding a new source.

### DIFF
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -121,6 +121,30 @@ class Source(models.Model):
         ('music_offtopic', 'Non-Music Section'),
     )
 
+    EMBED_METADATA_DEFAULT = bool(os.environ.get('EMBED_METADATA_DEFAULT', False))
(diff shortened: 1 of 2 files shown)


meeb:
Hi, thanks for the PR. While I can see the benefit of this for people that add a lot of sources I'm not sure this is the most sensible way. Adding 22 more ENV vars is pretty unwieldy. The way you've added them here is also problematic, changing the defaults on the model will create issues with creating migrations as the defaults are stored in the migrations. This also this introduces a scope for lots of errors if someone sets an ENV var to something not in a `choices` option, for example you set the resolution to `1080P` and not `1080p` or the video codec to `vp9` and not `VP9`.

Defaults are almost certainly better handled by `get_initial` for the `AddSourceView` here:

https://github.com/meeb/tubesync/blob/main/tubesync/sync/views.py#L371

which pre-populates the add source form on load.

rpowel:
Good points. I will close this PR and take another crack at it

meeb:
Thanks for taking the time to contribute!

PR #513, opened 2024-07-10
Use assertEqual as pluralized doesn't exist

joestump:
Was working on getting my local dev environment working and ran into this.


meeb:
Thanks! `assertEquals` did exist but seems it's been a depreciated alias since about 3.1 and removed in 3.6 or there about.

joestump:
@meeb ah, that makes sense. Could have sworn it existed, but it’d been a bit since I’d written Python tests in the Django framework. 😅

Issue #266, opened 2022-09-04
Add keyword and duration filters for sources


meeb:
Allow exclusion of media on a source by keyword or by length (to not download media under a certain length).

meeb:
Discussed in #265

timwhite:
Would also be good to exclude media over a certain length. One of my channels posts "marathons" which are a collection of previous videos all together. I really don't want those extra ones, especially as they are large!

timwhite:
@meeb Just letting you know that I'm working on this, and at the same time doing a cleanup up all the filtering into a single unified place. From my experience, it's best to express the filtering as "if any of these match, we skip it", which makes the logic easier to follow. I hope to have a branch for testing in the next few days, assuming I can get it to build for testing.

meeb:
Note regex title filtering was added by https://github.com/meeb/tubesync/pull/425

timwhite:
> Note regex title filtering was added by #425

Yep. I need duration filters. I'm also tempted to see if I can implement the last part of #425 that was discussed, so we can easily have the regex flip to an exclusion match instead of an inclusion match. That makes it much easier to write an exclusion list than a list of negative look aheads.

meeb:
Yep all sounds fine. I was linking that just PR on the off-chance you hadn't seen it so you weren't duplicating anything. Thanks for the effort!

Issue #509, opened 2024-06-24
Feature Request: More control through a Higher Order Function.


smolattack:
I see that there are a lot of feature requests that have various needs for filtering out unwanted videos. It would be a difficult and never-ending task to fulfill all such requests. Also, each channel/playlist will have it's own quirks, naming conventions and other patterns that might benefit from a bespoke filtering strategy. I have a proposal that might help alleviate this challenge.

## Phase 1
This is (hopefully) a quick and dirty way to get things started. Deep breath... How about `eval()`? OK, I know, lots of stigma and rightly so, for production code. This solution is for homelabs so it would allow for tinkering and you can put a big disclaimer in the form.

### Why?
The user gets more control without needing code changes. Also, this allows the community to help others by hand writing custom scripts for those that don't know python (like myself, but willing to learn if this feature is implemented).

The idea is that you would add a choice between a video title regex search pattern and a higher order function that gets called with `instance` as a parameter. If there is a way to also serialize and somehow output `instance`, that would be great (e.g., an ENV debug flag that would print a serialized `instance` in docker logs).

Something like this
Before:
https://github.com/meeb/tubesync/blob/e3b5d63501a0a8e2fffac64ee1a48b9a69421d3e/tubesync/sync/signals.py#L129

After:
```py
if instance.source.is_match(instance)
```
where `is_match` will be able to ascertain whether to `eval` and call the user defined function with `instance` or revert back to calling `is_regex_match` with the video title.

## Phase 2
To transition away from eval, you could ship https://nodered.org/ as a service and allow users a more visual way of providing bespoke filtering logic. Once integrated, this service could be used as an event emitter (e.g., start a workflow once a media file is downloaded, or an error occurs, etc...) and allow for even more control for the average homelabber's automation. For example, I could connect Node-RED a jellyfin HA integration via MQTT and automatically trigger library re-scans.

I think the features outlined above would have a very good value/effort ratio and would have compounding effects within the community, such that, members could more effectively help each other.

meeb:
I can see the logic behind the suggestion but I'd be very wary about adding a freeform text field that people can put Python code into that is `eval()`'d in-process. By default the tubesync has no authentication so the interface becomes a one-click RCE vector. If this ever existed it would be behind a `ENABLE_EVAL_YES_I_REALLY_KNOW_WHAT_IM_DOING=True` environment variable. If you know what you're doing you can always just use Docker volumes to spam in changes to signals anyway.

[...]

PR #515, opened 2024-07-13
Improve skip filtering with regex invert and duration limits

### LINKED ISSUE #482: Thumbnails downloading for videos older than download cap
I have YouTube sources added which contain thousands of videos. However, I only care about new videos, as I'm transitioning away from direct YouTube viewing, so I have the Download Cap set to 1 week for each source. I only care about videos from the last week, and new ones as they arrive.

However, TubeSync is still downloading thumbnails for EVERY video in each source. Looking at the queued tasks, there are thousands of tasks waiting to download very old thumbnails. It can take several hours for it to work through those and get to the actual videos. This also causes the database file to get very large.

Is this intended behavior, or can it be changed to entirely ignore downloads of content older than the download cap?

timwhite:
Fix #482, #266

- Allow filtering based on duration (either with a minimum or maximum duration)
- Allow filtering on regex and inverted regex (easier than lookahead regex's)
- Cleanup of all filtering code into single location (triggered on Media save)
- Bring title and duration into own fields in DB to allow filtering quickly without looking up metadata
- Update derived database fields from metadata on save

Source edit view
Screenshot from 2024-07-13 08-32-17

Source view (with inverted regex)
Screenshot from 2024-07-13 08-32-59


meeb:
This looks excellent, many thanks for the PR! I'll give your commits a once-over shortly.

meeb:
Zero issues obvious with your commits, nicely done. I'll merge it, give it a test and bundle it into the next release.

timwhite:
@meeb thanks. Happy to have had some time to contribute back. I have a QOL PR coming shortly, and also a way to solve #509 to think about

meeb:
@timwhite there might have been a couple of minor issues introduced with your filtering refactoring and issues with silly long titles. Feel free to review my two hot fix commits above (https://github.com/meeb/tubesync/commit/69a824e33403143f1289608dc37717c1dd62fded for #518 and https://github.com/meeb/tubesync/commit/68de80748344ea51eadb6320fcc3706cb65b2aaa for #519) if you get a moment.

timwhite:
@meeb those commits look good. I can see how the date bug came in. Interesting about the title length issue. All my googling said the maximum length was 100 characters. (e.g. https://developers.google.com/youtube/v3/docs/videos). I wonder if it's a utf encoding thing, but my understanding was that it was 100 utf characters, which is what the database was setup with. Your fixes look perfect.

meeb:
I'm pretty sure SQLite uses character counters and not byte counters so it's probably not that:

```sql
sqlite> select length('😀');
1
```

It's probably just YouTube returning massive titles. Either way, not a complex fix!

PR #516, opened 2024-07-13
Allow semi easy addition to filtering logic

### LINKED ISSUE #509: Feature Request: More control through a Higher Order Function.

timwhite:
Fix #509

Allow semi easy override of filtering logic with custom python code so it can be extended by end users

Have a look and let me know what you think. I'm not entirely sure it's needed, but it is a solution to #509

### DIFF
--- a/docs/custom-filters.md
+++ b/docs/custom-filters.md
@@ -1,0 +1,41 @@
+# TubeSync
+
+## Advanced usage guide - Writing Custom Filters
+
+Tubesync provides ways to filter media based on age, title string, and 
+duration. This is sufficient for most use cases, but there more complicated
+use cases that can't easily be anticipated. Custom filters allow you to 
+write some Python code to easily add your own logic into the filtering.
+
+Any call to an external API, or that requires access the metadata of the
+media item, will be much slower than the checks for title/age/duration. So
+this custom filter is only called if the other checks have already passed.
+You should also be aware that external API calls will significantly slow
+down the check process, and for large channels or databases this could be
+an issue.
+
+### How to use
+1. Copy `tubesync/sync/overrides/custom_filter.py` to your local computer
+2. Make your code changes to the `filter_custom` function in that file. Simply return `True` to skip downloading the item, and `False` to allow it to download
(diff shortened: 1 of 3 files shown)


meeb:
It's probably not needed but as a feature it's not intrusive and doesn't affect anything else so seems fine to me for the handful of users who might want to use something like this. Thanks for the PR!

PR #517, opened 2024-07-13
Docker build caching

timwhite:
Split installation of services, dependency installation, and code building steps so we can make the most use of docker build caches. When only making changes to code, this can now use the build cache to build an image in <10s. This means that if the base dependencies (redis/nginx etc) don't change between versions, the changed layers to download are less, allowing for quicker updates.

This also makes it easy to see runtime deps vs build only deps, so we can uninstall the packages we don't need once the pipenv is setup.

Dive report on ghcr.io/meeb/tubesync:latest with 10 layers
```
Total Image size: 621 MB             
Potential wasted space: 26 MB                  
Image efficiency score: 96 %
```

Dive report with these changes, with 12 layers, but quicker rebuild time, and also improved space efficiency
```
Total Image size: 603 MB                       
Potential wasted space: 8.1 MB         
Image efficiency score: 98 %
```

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -5,3 +5,4 @@
 README.md
 tubesync/media
 tubesync/downloads
+db.sqlite3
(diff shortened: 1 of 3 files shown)


meeb:
Appreciate the tidy-up!

PR #523, opened 2024-07-21
Add support for indexing Streams

FaySmash:
Tweaked the original code a little and added option to only index streams.


meeb:
Thanks! Looks reasonable. I'll give it a test shortly.

FaySmash:
Here
```
BooleanField(
        _('index videos'),
        default=False
```
the default should be set to `True` to keep the current behavior, an oversight.

FaySmash:
quick update: 90% of them are streams, still no issues

dls

meeb:
Thanks for the update. This has also been working fine for me in testing.

PR #537, opened 2024-08-24
Fix python version in manage.py

michalmiddleton:
```shell
root@tubesync-6c8bb86b79-s88cd:/app# /usr/bin/env python
/usr/bin/env: ‘python’: No such file or directory

root@tubesync-6c8bb86b79-s88cd:/app# /usr/bin/env python3
Python 3.11.2 (main, May  2 2024, 11:59:08) [GCC 12.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
```

### DIFF
--- a/tubesync/manage.py
+++ b/tubesync/manage.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 
 import os


meeb:
What distro are you using where `/usr/bin/env python` isn't available but `/usr/bin/env python3` is? Assuming this is Debian or Ubuntu based, do you have `python-is-python3` installed? I don't have a particular problem with this, but I'd need to check that `/usr/bin/env python3` is generally universally accepted or this will break things for other people.

meeb:
Had a check, seems `/usr/bin/env python3` is relatively generic now. Thanks.

michalmiddleton:
Hi @meeb , apologies for the lack of details. I run TubeSync in the container image you provide, which is Debian 12 I believe. 
TBH, it completely slipped my mind that somebody would run it in a non-containerized environment in these days :)
Thanks!

PR #556, opened 2024-11-19
Use YOUTUBE_DL_TEMPDIR setting

### LINKED ISSUE #218: Option to download file parts into a temporary directory for combination
When downloading files into a directory that's actively being monitored by a media server (I'm experiencing this with Jellyfin but I've seen this behavior with Plex as well), sometimes the media server will attempt to start indexing files before the video/audio combination has happened, which leads to the part files being locked so that they can't be renamed or deleted. This makes the downloader fail to finish the download even though it was actually successful - you usually end up with an audio.mkv, video.mkv, and combined.temp.mkv file.

If the audio and video file parts could be downloaded and combined in a temporary directory, and then moved into the monitored directory, this would no longer be a problem.

tcely:
Fixes #218

I took a look at implementing this while taking a break from the debugging. It wasn't too bad to start seeing the `.temp.` files in a directory Jellyfin doesn't monitor.


meeb:
Thanks! This looks fine to me. I assume you've tested it and it works as expected? If `YOUTUBE_DL_TEMPDIR` isn't set the previous behavior is maintained? I ask as in my brief testing some time ago output templates with paths was a little unpredictable.

Also the reason `DEBUG` wasn't passed on to `yt-dlp` was because if you leave this on and add a large channel the logs can quickly grow into the gigabytes which caused some complaints, it's probably fine to turn on though given the nature of what you should be using `DEBUG` for.

tcely:
> I assume you've tested it and it works as expected? If `YOUTUBE_DL_TEMPDIR` isn't set the previous behavior is maintained? I ask as in my brief testing some time ago output templates with paths was a little unpredictable.

I am running it now. It had downloaded all of the pieces to the `temp` directory (/download/cache, inside the container) as I wanted.

I even saw an error where a `.temp.` file was left behind as I was watching. Instead of marking the download as successful, as it did before, this time it retried the download as the final path wasn't found.

When it's not set, the key `temp` from `paths` is missing which leaves the `home` key set in `paths`, which appears to exactly match what `yt-dlp` did when using an absolute path for the output template.

> Also the reason `DEBUG` wasn't passed on to `yt-dlp` was because if you leave this on and add a large channel the logs can quickly grow into the gigabytes which caused some complaints, it's probably fine to turn on though given the nature of what you should be using `DEBUG` for.

That's understandable. Logs are handled nicely by docker in my environment.

It was quite frustrating, to me, that `TUBESYNC_DEBUG` didn't show what was being sent as arguments to the download program. I had considered another setting, but decided against that alternative.

meeb:
Sounds good to me! I'll merge this in and do a release shortly. I need to bump yt-dlp and ffmpeg anyway so it's a good time to add this as a feature. Thanks for your contribution, most appreciated.

PR #557, opened 2024-11-22
Configure minimum permitted video height with VIDEO_HEIGHT_CUTOFF env var

danielmawhirter:
"Get next best resolution or codec instead" has a minimum quality capped by `VIDEO_HEIGHT_CUTOFF` with no method to bypass it. This is undesirable when adding a Playlist source, where audio-only is not desired, but that would be the only way to trigger a download.

Allow video heights down to 120px to allow "next best" to fallback as far as necessary.

Possible better ways to resolve this:

- Have a configuration in the Source to define the lowest tolerable resolution (optionally allowing audio-only fallback)
- Expose a page in the webapp that allows global configuration, would require settings persistence to db

Both of these possibilities seem potentially desirable, but are much larger scope than this patch. If there's appetite, I may be willing to take on one of those projects.

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,13 +368,14 @@ useful if you are manually installing TubeSync in some other environment. These 
 | TUBESYNC_DEBUG              | Enable debugging                                             | True                                 |
 | TUBESYNC_WORKERS            | Number of background workers, default is 2, max allowed is 8 | 2                                    |
 | TUBESYNC_HOSTS              | Django's ALLOWED_HOSTS, defaults to `*`                      | tubesync.example.com,otherhost.com   |
-| TUBESYNC_RESET_DOWNLOAD_DIR | Toggle resetting `/downloads` permissions, defaults to True  | True
+| TUBESYNC_RESET_DOWNLOAD_DIR | Toggle resetting `/downloads` permissions, defaults to True  | True                                 |
 | GUNICORN_WORKERS            | Number of gunicorn workers to spawn                          | 3                                    |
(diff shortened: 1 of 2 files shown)


meeb:
Thanks for the PR. The current behavior is by design, without this limitation it causes (or used to at least) significant reports of terrible quality downloads when some media items default to 144p for some people in some situations. The "fall back to audio only if the video quality is terrible" option might work, however this could be more complex than it first appears as the download location would have to also dynamically change to avoid mixing media formats in single libraries for media servers.

The last time I looked at the 144p issue was a couple of years ago so the behavior and fallback of YouTube may have changed since then.

What specifically is your use case where downloading 144p media with tubesync? Are you saying you're interested in the audio only so don't care about the video quality? The core use case of tubesync is a PVR for YouTube so downloading 144p media is an edge case at best, failing downloads with only 144p streams was a preferable hack / compromise.

meeb:
Yeah I have less of an issue with that, it maintains the current behavior while allowing you to tweak it as needed. Just note that (at some point when I ever get the time) a settings page does appear it may replace this and you'll have to adjust your config. Can you quickly add this to the README as well please?

danielmawhirter:
Sounds like a good plan, and so added

meeb:
Thanks!

meeb:
@danielmawhirter note that I have renamed `VIDEO_HEIGHT_CUTOFF` to `TUBESYNC_VIDEO_HEIGHT_CUTOFF` to keep the tubesync related env vars named consistently as of https://github.com/meeb/tubesync/commit/2c12bf57971fca79df41d19fb6a770580f8ea72e

PR #558, opened 2024-11-22
Reworked Dockerfile

tcely:
Moved the hashes to build arguments.
Optimized the build by removing an unnecessary tar command.
Recorded the versions in the final container.
Used shell functions to remove repeated similar commands.
Used a copy of `/root` in `/tmp` instead of recreating it.
Cleaned up old unset variables.

I'm unsure why `ARCH` is being used, so I kept it. Using `TARGETARCH` instead makes more sense to me. Also, `ARCH44` was a complete mystery variable to me, so I removed it.

I'd suggest squashing these commits when you merge this. I was working through the website, so they are more like save points than discrete changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,10 @@ ARG TARGETARCH
 ARG TARGETPLATFORM
 
 ARG S6_VERSION="3.2.0.0"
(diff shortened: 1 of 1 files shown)


meeb:
Thanks! I didn't particularly have any issues with the old way and this seems a little over-engineered, but it's readable enough and moving the regularly updated hashes to args is indeed cleaner. Thanks for the tweaks.

PR #560, opened 2024-11-23
Spelled firewall correctly

tcely:
Anyone using this to turn off the healthcheck firewall needs to update the variable that they are using.

Maybe announce this in the UI if the setting is not true?


meeb:
Nice catch, thanks. I don't think this is used very much, it has no real security implications. If you're configuring an external health check it should be pretty obvious if it's configured incorrectly.

PR #562, opened 2024-11-25
Refinements for Dockerfile

tcely:
The most important is catching when the `s6` or `ffmpeg` binaries aren't actually working correctly.


meeb:
What build system are you using? I don't particularly mind merging these after looking at them, but a lot of them seem to be accounting for what look like largely impossible failures or dumping info that's likely not too useful.

Is there some use case for all of this?

tcely:
Actually, the inspiration was from seeing the logs from the GitHub workflow.

I was surprised that the `arm64` build wasn't failed before you merged the removal of `=amd64` from `TARGETARCH` so I wanted to prevent a push of a container with incompatible `ffmpeg` from going unnoticed.

The change of the scope for when `HOME` was altered should not make a difference now, but if a later addition didn't account for that, it might cause confusion later. It's just better practice to limit which commands see that altered value, to as few as we need.

The `mkdir` output did answer a question I had about if some of the directories were existing already. The `file` output would have allowed me to see the type of binary that was extracted rather than inferring the problem in the older logs through the error or URL that was logged.

meeb:
Far enough, no real objections to your adjustments, I'll merge them into the next release.

tcely:
Thank you for your great software and prompt reviews!

PR #565, opened 2024-12-02
Sort formats by height and id

tcely:


tcely:
Closed because my test branch was merged into the same branch.

PR #566, opened 2024-12-02
Sort video formats by height and bitrate

tcely:
* Stop sorting audio formats
The list we received from `youtube-dl` is better than bitrate sorting

* Include candidates by height, not format (`format_note`)
* Populate the empty format when the height matched
* Match `VP09` to `VP9` to avoid unnecessary fallback

### DIFF
--- a/tubesync/sync/utils.py
+++ b/tubesync/sync/utils.py
@@ -1,6 +1,7 @@
 import os
 import re
 import math
+from operator import itemgetter
 from pathlib import Path
 import requests
 from PIL import Image
@@ -132,6 +133,15 @@ def seconds_to_timestr(seconds):
    minutes = seconds // 60
    seconds %= 60
    return '{:02d}:{:02d}:{:02d}'.format(hour, minutes, seconds)
+
+
+def multi_key_sort(sort_dict, specs, use_reversed=False):
+    result = list(sort_dict)
+    for key, reverse in reversed(specs):
+        result = sorted(result, key=itemgetter(key), reverse=reverse)
+    if use_reversed:
+        return list(reversed(result))
+    return result
 
 
 def parse_media_format(format_dict):
--- a/tubesync/sync/utils.py
+++ b/tubesync/sync/utils.py
@@ -158,6 +158,8 @@ def parse_media_format(format_dict):
         vcodec = None
     if vcodec == 'NONE':
         vcodec = None
+    if vcodec == 'VP09':
+        vcodec = 'VP9'
     acodec_full = format_dict.get('acodec', '')
     acodec_parts = acodec_full.split('.')
     if len(acodec_parts) > 0:
--- a/tubesync/sync/matching.py
+++ b/tubesync/sync/matching.py
@@ -87,7 +87,7 @@ def get_best_video_format(media):
         return False, False
     # Filter video-only formats by resolution that matches the source
     video_formats = []
-    sort_keys = [('height', False), ('id', False)] # key, reverse
+    sort_keys = [('height', False), ('vcodec', True), ('vbr', False)] # key, reverse
     for fmt in media.iter_formats():
         # If the format has an audio stream, skip it
         if fmt['acodec'] is not None:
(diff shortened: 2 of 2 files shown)


meeb:
Thanks! Just to clarify can you write something as to why it's beneficial to support additional ordering of media stream matches by ID?

tcely:
Yeah, when I actually get this figured out I'll write up more.

The idea here was to pick the best quality when the fallback function is used. I noticed that some of my videos had at least four options and one of the lower bitrates is being matched first.

meeb:
The format matching was quite carefully chosen, the quality is also not determined by the order of the stream ID either (special streams  (live, dash, overlay subtitles etc.) can have weird IDs not in sequence last time I looked.

These are more general questions, for example if you select h264 1080p but there's only h264 720p available but there's also an AV1 1080p stream should that replace it? Generally until now the position has been yes, probably.

I'm unlikely to merge something that orders based on an arbitrary ID, you're welcome to submit a PR for ordering based on known metadata though (video resolution, codecs, HDR flags, etc.).

tcely:
What can you tell me about this:

```
	
ID: 616 PREMIUM 616 - 1920x1080 (Premium) , PREMIUM (1920x1080), fps:30.0, video:VP09 @5711k
ID: 614 614 - 1920x1080 , (1920x1080), fps:30.0, video:VP09 @3246k
ID: 399 1080P 399 - 1920x1080 (1080p) , 1080P (1920x1080), fps:30, video:AV01 @487k
ID: 270 270 - 1920x1080 , (1920x1080), fps:30.0, video:AVC1 @4694k
ID: 248 1080P 248 - 1920x1080 (1080p) , 1080P (1920x1080), fps:30, video:VP09 @515k
ID: 137 1080P 137 - 1920x1080 (1080p) , 1080P (1920x1080), fps:30, video:AVC1 @629k (matched)
```

From what I have discovered 270 is not being considered because it doesn't have a note to match against the source format.

I had initially believed that 137 was just matched first, hence the idea about sorting by ID when the height was the same.

Also, I will likely look into avoiding drc audio, with perhaps an option to control that.

meeb:
What do you mean by note? 137 has a higher bitrate, also is 137 a combined audio and video stream? DRC streams seem new, no objection to adding a "DRC yes/no" flag.

tcely:
> What do you mean by note?

```
format_dict.get('format_note', '')
```

The `format` key is being sourced from `format_note` unless I missed something.

The bitrates for 270 is about 4,000k higher than for 137. No, 137 is a video only stream.

meeb:
Ah right yes, that note. Yeah I was reading on my phone and missed the other digit on the bitrate.

tcely:
No problem. I'm currently running these changes and things look good.

Let me know if you see anything that needs to be adjusted.

tubesync/sync/utils.py
```diff
@@ -148,6 +158,8 @@ def parse_media_format(format_dict):
         vcodec = None
     if vcodec == 'NONE':
         vcodec = None
+    if vcodec == 'VP09':
+        vcodec = 'VP9'
```
tcely: Remove this when #571 has been merged.

meeb:
The only thing I'm not entirely sold on here is swapping bitrate ordering of the streams to just whatever arbitrary order the YouTube interface labels with IDs. The (far distant) plan is to add generic support for sites to tubesync (any site `yt-dlp` supports) so relying on something YouTube specific might be a bit restrictive.

For example with audio, why would you not want the stream with the matching codec but highest quality bitrate?

tcely:
It's not YouTube providing the ordering, it's `yt-dlp` doing that.

I think in the future we may want to use `-S` more than `-f` because I just ran into a problem with the matched format not being available. With `yt-dlp -S res:480` it would have just tried another format instead of failing repeatedly.

The quick way past that problem, for now, was to put `check_formats` in the `get_media_info` options.

The highest quality bitrate audio was the drc when we did the sorting. The quality selector for `yt-dlp` promotes the normal audio higher consistently. I didn't think the simple bitrate sorting was better than what we were already provided for the audio.

meeb:
I must admit I've not looked into this much since I hacked up the first version of tubesync other than tweaks, I probably need to revisit some of the yt-dlp internals again. That makes sense. Part of why tubesync has the somewhat over the top matching was to work around that you couldn't (previously at least) easily pick and reliably get the codec, resolution and other features you wanted for each stream with sensible fallbacks, similar to how other PVR's work (e.g. radarr).

I'm fully open to falling back to yt-dlp's handling if this is sufficiently reliable, obviously yt-dlp upstream has a much wider developer support base so using whatever they provide would be preferable.

PR #569, opened 2024-12-05
Adjust to correct 60 fps case

tcely:
This block is in the not 60 fps case, so checking for a match outside of that condition doesn't make sense to me or fit the comment.


tcely:
@meeb Can this be merged or is there something intentional about that condition?

meeb:
No reason, I've just been busy this week and not reviewed it. Thanks for the PR!

PR #571, opened 2024-12-05
Handle STR0NNN in a generic way

tcely:
The function should return 'VP9' for:

- vp9.1.2.3
- vp09.4.5.6

As well as other variations.
It also works for avc01 or av01, or any others that fit the pattern.

### DIFF
--- a/tubesync/sync/utils.py
+++ b/tubesync/sync/utils.py
@@ -134,6 +134,21 @@ def seconds_to_timestr(seconds):
    return '{:02d}:{:02d}:{:02d}'.format(hour, minutes, seconds)
 
 
+def normalize_codec(codec_str):
+    result = str(codec_str)
+    parts = result.split('.')
+    if len(parts) > 0:
+        result = parts[0].strip()
+    else:
+        return None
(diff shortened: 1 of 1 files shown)


tcely:
I've been running with this code for a while now. I'm seeing more exact matches than before.

meeb:
This looks fine, but preferably could do with some tests for the new function.

tcely:
The matching tests are already exercising that function.

Was there something specific you were looking to test?

meeb:
Ah, I just meant as you're (slightly) modifying how the codec string was being parsed if there was a test that explicitly confirmed this behavior. Current tests are probably sufficient.

tcely:
I'm up for adding more tests, especially as the boring metadata doesn't have any `vcodec` matching this `vp09.01.2.40` pattern, that I saw in my database.

The 20230629 metadata does have both `vp09` and `av01` codecs though.

meeb:
Yeah YouTube especially seem to tweak the codecs quite frequently. The version number postfixes are a new-ish change.

PR #572, opened 2024-12-09
Add Support for TUBESYNC_DIRECTORY_PREFIX Environment Variable

jameswlane:
### **Description**
This PR introduces a new environment variable, `TUBESYNC_DIRECTORY_PREFIX`, to enhance the flexibility of download directory configurations in TubeSync. Additionally, it updates the relevant documentation and tests to reflect this new feature.

---

### **Key Changes**
1. **New Feature**:
   - Added `type_directory_path` property to support two directory modes:
     - `default`: Keeps the original behavior where audio and video files are stored in their respective directories.
     - `flat`: Places all files in the root directory.

2. **Validation**:
   - Added validation for incorrect or malformed `TUBESYNC_DIRECTORY_PREFIX` values to prevent runtime errors.

3. **Tests**:
   - Created tests for the `type_directory_path` method, covering:
     - Default behavior.
     - Flat directory mode.

4. **Documentation**:
   - Updated the README to include `TUBESYNC_DIRECTORY_PREFIX` under the list of environment variables.
   - Added usage examples and possible values.

---

### **Backward Compatibility**
- The `default` mode maintains the existing behavior, ensuring backward compatibility for users not setting this environment variable.

---

### **Checklist**
- [x] Code changes tested locally.
- [x] Unit tests added for new functionality.
- [x] Documentation updated.
- [x] Backward compatibility verified.

---

### **Additional Notes**
Feedback is welcome, especially on:
- Edge case handling for custom directory mode.
- Any additional test cases that could improve coverage.

---

### DIFF
--- a/tubesync/sync/tests.py
+++ b/tubesync/sync/tests.py
@@ -6,7 +6,9 @@
 
 
 import logging
+import os
 from datetime import datetime, timedelta
+from pathlib import Path
 from urllib.parse import urlsplit
 from xml.etree import ElementTree
 from django.conf import settings
@@ -1714,3 +1716,77 @@ class TasksTestCase(TestCase):
         self.assertEqual(src1.media_source.all().count(), 3)
         self.assertEqual(src2.media_source.all().count(), 2)
         self.assertEqual(Media.objects.filter(pk=m22.pk).exists(), False)
+
+class TypeDirectoryPathTestCase(TestCase):
+    def setUp(self):
+        # Mock settings for testing
+        self.audio_dir = Path("/mock/audio/dir")
+        self.video_dir = Path("/mock/video/dir")
+        self.download_dir = Path("/mock/download/dir")
+        settings.DOWNLOAD_AUDIO_DIR = self.audio_dir
+        settings.DOWNLOAD_VIDEO_DIR = self.video_dir
+        settings.DOWNLOAD_DIR = self.download_dir
+
+        # Create a source object for testing
+        self.source = Source(
+            directory="test_directory",
+            source_resolution=Source.SOURCE_RESOLUTION_AUDIO,
+        )
+
+    def test_default_mode_audio(self):
+        """
+        Test default mode for audio resolution.
+        """
+        os.environ["TUBESYNC_DIRECTORY_MODE"] = "default"
+        expected_path = self.audio_dir / self.source.directory
+        self.assertEqual(self.source.type_directory_path, expected_path)
+
+    def test_default_mode_video(self):
+        """
+        Test default mode for video resolution.
+        """
+        os.environ["TUBESYNC_DIRECTORY_MODE"] = "default"
+        self.source.source_resolution = Source.SOURCE_RESOLUTION_1080P
+        expected_path = self.video_dir / self.source.directory
+        self.assertEqual(self.source.type_directory_path, expected_path)
+
+    def test_flat_mode(self):
+        """
+        Test flat mode places files in the root download directory.
+        """
+        os.environ["TUBESYNC_DIRECTORY_MODE"] = "flat"
+        expected_path = self.download_dir / self.source.directory
+        self.assertEqual(self.source.type_directory_path, expected_path)
+
+    def test_custom_mode_audio(self):
+        """
+        Test custom mode with prefixes for audio.
+        """
+        os.environ["TUBESYNC_DIRECTORY_MODE"] = "custom:audio_prefix,video_prefix"
+        expected_path = self.download_dir / "audio_prefix" / self.source.directory
+        self.assertEqual(self.source.type_directory_path, expected_path)
+
+    def test_custom_mode_video(self):
+        """
+        Test custom mode with prefixes for video.
+        """
+        os.environ["TUBESYNC_DIRECTORY_MODE"] = "custom:audio_prefix,video_prefix"
+        self.source.source_resolution = Source.SOURCE_RESOLUTION_1080P
+        expected_path = self.download_dir / "video_prefix" / self.source.directory
+        self.assertEqual(self.source.type_directory_path, expected_path)
+
+    def test_custom_mode_invalid_format(self):
+        """
+        Test custom mode with an invalid format.
+        """
+        os.environ["TUBESYNC_DIRECTORY_MODE"] = "custom:only_audio_prefix"
+        with self.assertRaises(ValueError):
+            _ = self.source.type_directory_path
+
+    def test_invalid_mode(self):
+        """
+        Test unsupported directory mode raises an error.
+        """
+        os.environ["TUBESYNC_DIRECTORY_MODE"] = "unsupported_mode"
+        with self.assertRaises(ValueError):
+            _ = self.source.type_directory_path
--- a/tubesync/sync/tests.py
+++ b/tubesync/sync/tests.py
@@ -1719,15 +1719,9 @@ class TasksTestCase(TestCase):
 
 class TypeDirectoryPathTestCase(TestCase):
     def setUp(self):
-        # Mock settings for testing
-        self.audio_dir = Path("/mock/audio/dir")
-        self.video_dir = Path("/mock/video/dir")
-        self.download_dir = Path("/mock/download/dir")
-        settings.DOWNLOAD_AUDIO_DIR = self.audio_dir
-        settings.DOWNLOAD_VIDEO_DIR = self.video_dir
-        settings.DOWNLOAD_DIR = self.download_dir
-
-        # Create a source object for testing
+        self.audio_dir = Path(settings.DOWNLOAD_AUDIO_DIR)
+        self.video_dir = Path(settings.DOWNLOAD_VIDEO_DIR)
+        self.download_dir = Path(settings.DOWNLOAD_ROOT)
         self.source = Source(
             directory="test_directory",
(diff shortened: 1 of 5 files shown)


tcely:
Why are we modifying code in `models.py` at all?

Changing the values of the settings would seem like a much cleaner approach.

jameswlane:
@tcely I will take a look at this when I get back to my desk. Thank you for the suggestion. 🙏

tubesync/sync/tests.py
```diff
@@ -1714,3 +1716,77 @@ def test_delete_old_media(self):
...
+
+class TypeDirectoryPathTestCase(TestCase):
+    def setUp(self):
+        # Mock settings for testing
+        self.audio_dir = Path("/mock/audio/dir")
+        self.video_dir = Path("/mock/video/dir")
+        self.download_dir = Path("/mock/download/dir")
+        settings.DOWNLOAD_AUDIO_DIR = self.audio_dir
+        settings.DOWNLOAD_VIDEO_DIR = self.video_dir
+        settings.DOWNLOAD_DIR = self.download_dir
```
tcely: Don't change settings directly.

DOWNLOAD_ROOT is the path to change anyway.

tubesync/sync/tests.py
```diff
@@ -1714,3 +1716,77 @@ def test_delete_old_media(self):
         self.assertEqual(src1.media_source.all().count(), 3)
         self.assertEqual(src2.media_source.all().count(), 2)
         self.assertEqual(Media.objects.filter(pk=m22.pk).exists(), False)
+
+class TypeDirectoryPathTestCase(TestCase):
+    def setUp(self):
+        # Mock settings for testing
+        self.audio_dir = Path("/mock/audio/dir")
+        self.video_dir = Path("/mock/video/dir")
+        self.download_dir = Path("/mock/download/dir")
```
tcely:
```
self.download_dir = settings.DOWNLOAD_ROOT
```

tubesync/sync/tests.py
```diff
@@ -1714,3 +1716,77 @@ def test_delete_old_media(self):
...
+            directory="test_directory",
+            source_resolution=Source.SOURCE_RESOLUTION_AUDIO,
+        )
+
+    def test_default_mode_audio(self):
+        """
+        Test default mode for audio resolution.
+        """
+        os.environ["TUBESYNC_DIRECTORY_MODE"] = "default"
+        expected_path = self.audio_dir / self.source.directory
```
tcely:
```
expected_path = self.download_dir / "audio" / self.source.directory
```

tubesync/sync/tests.py
```diff
@@ -1714,3 +1716,77 @@ def test_delete_old_media(self):
...
+        self.assertEqual(self.source.type_directory_path, expected_path)
+
+    def test_default_mode_video(self):
+        """
+        Test default mode for video resolution.
+        """
+        os.environ["TUBESYNC_DIRECTORY_MODE"] = "default"
+        self.source.source_resolution = Source.SOURCE_RESOLUTION_1080P
+        expected_path = self.video_dir / self.source.directory
```
tcely:
```
expected_path = self.download_dir / "video" / self.source.directory
```

tcely (changes requested):
Changes needed for accurate testing.

jameswlane:
With the new changes made there are some broken test, I got two back in working order but the other five are eluding me. I will pick up debugging in the morning.

tubesync/sync/tests.py
```diff
@@ -1719,15 +1719,9 @@ def test_delete_old_media(self):
...
-        self.audio_dir = Path("/mock/audio/dir")
-        self.video_dir = Path("/mock/video/dir")
-        self.download_dir = Path("/mock/download/dir")
-        settings.DOWNLOAD_AUDIO_DIR = self.audio_dir
-        settings.DOWNLOAD_VIDEO_DIR = self.video_dir
-        settings.DOWNLOAD_DIR = self.download_dir
-
-        # Create a source object for testing
+        self.audio_dir = Path(settings.DOWNLOAD_AUDIO_DIR)
+        self.video_dir = Path(settings.DOWNLOAD_VIDEO_DIR)
```
tcely: You won't want to set `audio_dir` and `video_dir` in your tests at all.
These are the values being tested, after all.

tubesync/sync/tests.py
```diff
@@ -1719,15 +1719,9 @@ def test_delete_old_media(self):
...
+        self.download_dir = Path(settings.DOWNLOAD_ROOT)
```
tcely: I believe `Path` is not needed here because it is already a `Path` instance.

meeb:
Thanks for the thorough go at this feature @jameswlane ! Fetching this in `settings.py` is indeed the correct place with Django to toggle this feature.

Just to add to the feedback, how likely is the `custom:` prefix likely to be used? Currently this allows weird things like path traversal with the prefix, while almost certainly not a security issue given it'll have to be set on the host anyway it's probably not going to function as intended if anyone puts any dots in the field. It's also getting a bit overly "magic".

Would a `TUBESYNC_DIRECTORY_PREFIX` as a `True/False` defaulting to `True` work well enough here for your use case? It would certainly be less hacky to implement and just a single `if` toggle to enable.

tcely:
> The main issue I’m facing is that updating the environmental variable doesn’t reflect in the settings when testing. I haven’t found a way to reload the settings file to pick up the new variable.

There is also [modify_settings](https://docs.djangoproject.com/en/5.1/topics/testing/tools/#django.test.SimpleTestCase.modify_settings) but I don't think that solves the issue you are having.

As much as I dislike duplicated code, it might just be easier to copy the logic from settings into the test cases. I'd like to see the tests actually use the filename attribute from Media to compare to expected values.

tubesync/sync/tests.py
```diff
@@ -1719,68 +1719,28 @@ def test_delete_old_media(self):
...
-    def test_default_mode_audio(self):
+    def test_directory_prefix_default(self):
         """
-        Test default mode for audio resolution.
+        Test that default directory prefix exist.
         """
-        os.environ["TUBESYNC_DIRECTORY_MODE"] = "default"
-        expected_path = self.audio_dir / self.source.directory
-        self.assertEqual(self.source.type_directory_path, expected_path)
+        os.environ['TUBESYNC_DIRECTORY_PREFIX'] = ''
```
tcely, line 1731: Is this needed for the default case? In other words, we should already be in the default scenario just by loading settings.

meeb:
@jameswlane you can totally ignore environment variables in tests. Your test should be on the `settings.TUBESYNC_DIRECTORY_PREFIX` setting which should be either `True` or `False` and the changes in behavior this flag toggles, that's about it. The env var stuff is different, that's adjusting Django settings from the container environment so it can be deployed in a container, that's different from the actual feature. You really just need something like:

```python
def test_directory_prefix(self):
    # Confirm the setting exists and is valid
    self.assertTrue(hasattr(settings, 'TUBESYNC_DIRECTORY_PREFIX'))
    self.assertTrue(isinstance(settings.TUBESYNC_DIRECTORY_PREFIX))
    # Test the default behavior for "True", forced "audio" or "video" parent directories for sources
    settings.TUBESYNC_DIRECTORY_PREFIX = True
    self.source.source_resolution = Source.SOURCE_RESOLUTION_AUDIO
    test_audio_prefix_path = Path(self.source.directory_path)
    self.assertEqual(test_audio_prefix_path.parts[-2], 'testdirectory')
    self.assertEqual(test_audio_prefix_path.name[-1], 'audio')
    self.source.source_resolution = Source.SOURCE_RESOLUTION_1080P
    test_video_prefix_path = Path(self.source.directory_path)
    self.assertEqual(test_video_prefix_path.parts[-2], 'testdirectory')
    self.assertEqual(test_video_prefix_path.parts[-1], 'video')
    # Test the default behavior for "False", no parent directories for sources
    settings.TUBESYNC_DIRECTORY_PREFIX = True
    test_no_prefix_path = Path(self.source.directory_path)
    self.assertEqual(test_no_prefix_path.parts[-1], 'testdirectory')
```
  
  You don't need to make this its own test unit class, you can just include it in `FilepathTestCase` which already creates a test `self.source` you can test on..

Also you should just add `TUBESYNC_DIRECTORY_PREFIX = True` to `settings.py` then you override it with the`os.getenv()` stufff in `local_settings.py.container`.

Basically only the container build should care about env vars (other than maybe `DEBUG`).

meeb:
I'm being pretty pedantic for this little change, I'll merge this and quickly patch it myself. Thanks for the contribution @jameswlane - I'll link back in this thread what I tweak just so you can see the changes if you're interested.

meeb:
@jameswlane you can see what I've adjusted in https://github.com/meeb/tubesync/commit/5f28b8be661849071a54190b84062d562bb20abd

tcely:
@meeb What happens with your code when you have 'true' or 'false' as strings?

meeb:
Oh shush :) https://github.com/meeb/tubesync/commit/6c12244a2521ad954b078871a107da1a3cdac66d

jameswlane:
@meeb @tcely Thank you for your help! It’s been quite a few years since I last worked with Python and Django—probably back in the 2.x days around 2018—so this has been a great refresher. I also want to apologize for needing so much guidance along the way.

PR #573, opened 2024-12-09
Accept '-' in sub_langs

### LINKED ISSUE #554: Subtitles with hyphen.
I'm trying to download some videos with portuguese Brazilian subtitles. In yt-dlp they are classified as "pt-BR" subtitle, but when I write this on tubesync, i always have the error "Subtitle langs must be a comma-separated list of langs. example: en,fr or all,-fr,-live_chat".
I tested directly on yt-dlp, and works perfectly in the command-line writing "pt-BR".
image
image
image

tcely:
Fixes #554

### DIFF
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -422,7 +422,7 @@ class Source(models.Model):
         help_text=_('List of subtitles langs to download, comma-separated. Example: en,fr or all,-fr,-live_chat'),
         validators=[
             RegexValidator(
-                regex=r"^(\-?[\_\.a-zA-Z]+,)*(\-?[\_\.a-zA-Z]+){1}$",
+                regex=r"^(\-?[\_\.a-zA-Z-]+,)*(\-?[\_\.a-zA-Z-]+){1}$",
                 message=_('Subtitle langs must be a comma-separated list of langs. example: en,fr or all,-fr,-live_chat')
             )
         ]


tcely:
@meeb While looking into this, I was left wondering why it's using `A*A{1}` as a construction instead of simply `A+` as was used for the internal characters.

tubesync/sync/models.py
```diff
@@ -422,7 +422,7 @@ class IndexSchedule(models.IntegerChoices):
         help_text=_('List of subtitles langs to download, comma-separated. Example: en,fr or all,-fr,-live_chat'),
         validators=[
             RegexValidator(
-                regex=r"^(\-?[\_\.a-zA-Z]+,)*(\-?[\_\.a-zA-Z]+){1}$",
```
tcely, line 425: @meeb While looking into this, I was left wondering why it's using `^(A,)*A{1}$` as a construction instead of simply `^(A(,|$))+` to avoid repetition.

meeb: This was from a previous PR, I'm fine with this being corrected. Thanks!

meeb:
Cheers!

tubesync/sync/models.py
tcely: Alright. The proposed change is in #574.

PR #574, opened 2024-12-10
Simplify regex for sub_langs

tcely:


meeb:
Looking at the original regex I assume it was an attempt to allow something like `blah` or `blah-blah` but not `blah-blah-blah`? Obviously didn't validate much so simplifying it is fine. Cheers!

tcely:
When I first saw it I was confused, but I think it was only an attempt to allow `A,` repeated, or absent, to end with `A$` which sort of makes sense.

The alternation should achieve the same though without repeating the `A` group.

tcely:
Related to this change I think there may be a new migrations file needed. I'm not sure how that all functions though or why the regex changes would cause that.

Issue #570, opened 2024-12-05
Download Cap & Delete Old Media/Days to Keep conflict


Nanganator:
It seems like the two items mentioned in title conflict (perhaps just my interpretation).

My understanding of Download Cap is to have tubesync NOT download videos which were posted by the uploader beyond the set date.
My understanding of Delete Old Media/Days to Keep is to delete video files which were downloaded more than “Days to keep” days ago. 

However it seems when monitoring a playlist, even with Download Cap set to set to No Cap, if Days to Keep is set to 14, and Delete Old Media is ticked, items added to the playlist which were posted by the uploaded more than 14 days ago will not be downloaded.

tcely:
The situation is a bit complicated. Here is my understanding:

* `Download cap` is to prevent downloading items published before that time (I use 7 days for news channels).
* `Delete old media` (I use 10 days for news channels) is when the media item is removed from the TubeSync database.
After it's removed, a later index of the channel is likely to add it back but it won't download again because it is now older than the download cap.
* `Delete removed media` is for removing items that are no longer in the source channel or playlist from the TubeSync database.
* `Delete files on disk` should control if the downloaded media file is deleted from the disk when anything is removed from the TubeSync database.

I've personally seen my files on disk survive beyond 10 days, so I'm not sure how well that last option is currently working. However, it's better to fail to delete files than to delete too many files. 🙃

Nanganator:
Yes this is exactly as I have mentioned above so we are both expecting the same behaviour.

Issue I’m having is setting `Download Cap` is not working as you and I both expect it.
I have a playlist as a media source. `Download Cap` is set to `No Cap`. I expect this to mean that any video added to the playlist, regardless of date uploaded, should download. Unfortunately this does not happen.
I also have `Delete Old Media` enabled and `Days to Keep` set to `14 Days`. Unfortunately what this does is stops video’s which were uploaded more than 14 days ago to NOT be downloaded. Whereas I would expect this to mean media which was downloaded by TubeSync more than 14 days ago to be deleted.

So as I mentioned, it seems to be some kind of conflict or wrong logic with these settings.

tcely:
> I have a playlist as a media source. `Download Cap` is set to `No Cap`. I expect this to mean that any video added to the playlist, regardless of date uploaded, should download. Unfortunately this does not happen.

I have a similar playlist. I do see downloads happening from that playlist even for videos that were uploaded many years ago.

[...]

PR #576, opened 2024-12-11
Disable checking against days to keep

tcely:
This isn't the logic we should be using.
Just stop using it until the function is comparing the proper dates.

See: #570


meeb:
Yep seems reasonable.

PR #577, opened 2024-12-11
Adjust days_to_keep logic

### LINKED ISSUE #570: Download Cap & Delete Old Media/Days to Keep conflict

tcely:
Fixes #570

### DIFF
--- a/tubesync/sync/filtering.py
+++ b/tubesync/sync/filtering.py
@@ -23,8 +23,8 @@ def filter_media(instance: Media):
         skip = True
 
     # Check if older than source_cutoff
-    #if filter_source_cutoff(instance):
-    #    skip = True
+    if filter_source_cutoff(instance):
+        skip = True
 
     # Check if we have filter_text and filter text matches
     if filter_filter_text(instance):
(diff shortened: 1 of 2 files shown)


tubesync/sync/filtering.py
```diff
@@ -127,20 +127,15 @@ def filter_max_cap(instance: Media):
...
-            return True
+    if instance.source.delete_old_media and instance.source.days_to_keep_date:
+        if not instance.downloaded or not isinstance(instance.download_date, datetime):
+            return False
 
-        delta = timezone.now() - timedelta(days=instance.source.days_to_keep)
-        if instance.published < delta:
-            # Media was published after the cutoff date, skip it
+        days_to_keep_age = instance.source.days_to_keep_date
+        if instance.download_date < days_to_keep_age:
```
meeb, line 137: This can do a `datetime.datetime < False` comparison and raise a `TypeError` if `source.days_to_keep <= 0` I think? Just needs an `if days_to_keep_age and ...` check.

tcely: Check the first if conditions.

We could check it again, or comment the code about why it won't be `False` to avoid confusion by later readers.

meeb: Oh right you evaluate it twice, fair enough. Reviewing code on a phone is hard.

meeb:
Thanks!

PR #578, opened 2024-12-11
Remove audio / video directory modification

tcely:
This was from #572 but isn't needed with the later modifications.


meeb:
Oh derp, I removed that then undid it. Cheers.

PR #579, opened 2024-12-11
Simplify directory_path

tcely:
This looked to be identical code to what `filename` used. Remove the duplicated code by using the function instead.


meeb:
Yep! There's a load of duplication in some areas. Also the models being a couple of massive classes is pretty messy. Throwbacks to when tubesync was some hacky nonsense before I released it.

PR #580, opened 2024-12-11
Report db.sqlite3 size on dashboard

tcely:


meeb:
No objection to this, but probably might be nicer to use the database connection and check the vendor that's being used:

```python
>>> from django.db import connection
>>> connection.vendor
'sqlite'
```

Using the database from the connection parameters might be cleaner too:

```
>>> connection.get_connection_params()['database']
'/home/some/path/to/a/db.sqlite3'
```

tcely:
Those are good suggestions. I'll look into improving this.

meeb:
Looks nicer, you happy for this to be merged?

tcely:
Yeah. filesizeformat needs to be fixed to use proper prefixes, but that can happen later.

PR #581, opened 2024-12-11
SQLite connection options

tcely:
Configure the connection `PRAGMA` values from settings.

It's now less messy.
However, the `DatabaseWrapper` is not exactly fun to work with.

I've made changes to hopefully make it as easy to maintain as I can.


tcely:
@meeb Is there a good reason Django is pinned at `3.2`?

meeb:
3.2 is the latest version that works with the background tasks module which is basically now unmaintained. Upgrading past 3.2 requires replacing the tasks system.

meeb:
I'd much rather wait for the tasks system to be worked on and Django upgraded to 5.1 to natively support these sorts of features rather than backport this.

tcely:
> Upgrading past 3.2 requires replacing the tasks system.

Did you have something in mind for a replacement?

[APScheduler](https://pypi.python.org/pypi/APScheduler/) looked interesting, but I haven't used it before.

meeb:
I was porting it to Celery at one point, however that got delayed as I started waiting for native background workers (Django DEP14) to merge into main.

tcely:
Are you planning to wait for the implementation to be completed?

It looks like there is enough support for redis or sqlite task storage already.

https://github.com/RealOrangeOne/django-tasks

meeb:
Yeah I didn't particularly want to implement it against an in-flux reference implementation then have to make further changes once it was mainline.

meeb:
Is this still needed in this form after the bump to Django 5.X ?

tcely:
The settings are going to be the same but the rest is probably unnecessary now.

tcely:
We should both double check these settings are what Django 5.1 supports.

Once we are sure it works with them, this can be merged. It should help with locking problems too.

tubesync/tubesync/local_settings.py.container
```diff
@@ -46,6 +46,14 @@ else:
         'default': {
             'ENGINE': 'django.db.backends.sqlite3',
             'NAME': CONFIG_BASE_DIR / 'db.sqlite3',
+            "OPTIONS": {
+                "transaction_mode": "IMMEDIATE",
```
tcely, line 50: https://docs.djangoproject.com/en/5.1/ref/databases/#transactions-behavior

tubesync/tubesync/local_settings.py.container
```diff
@@ -46,6 +46,14 @@ else:
...
+                "init_command": """
+                    PRAGMA legacy_alter_table = OFF;
+                    PRAGMA auto_vacuum = INCREMENTAL;
+                    PRAGMA incremental_vacuum(100);
+                """,
+            },
```
tcely, line 56: https://docs.djangoproject.com/en/5.1/ref/databases/#setting-pragma-options

tubesync/tubesync/local_settings.py.container
```diff
@@ -46,6 +46,14 @@ else:
         'default': {
             'ENGINE': 'django.db.backends.sqlite3',
             'NAME': CONFIG_BASE_DIR / 'db.sqlite3',
+            "OPTIONS": {
+                "transaction_mode": "IMMEDIATE",
+                "init_command": """
+                    PRAGMA legacy_alter_table = OFF;
+                    PRAGMA auto_vacuum = INCREMENTAL;
+                    PRAGMA incremental_vacuum(100);
```
tcely, line 54: https://docs.djangoproject.com/en/5.1/ref/databases/#setting-pragma-options

meeb:
Yeah those options should now be natively supported.

tcely:
I concur.

It's ready to merge. 🎉

PR #587, opened 2024-12-20
Use tempfile for write_text_file

tcely:
This is more atomic and allows for checking `filepath` with `file_is_editable` before replacing the original file.


meeb:
Thanks!

PR #589, opened 2024-12-20
Remove forcejson option

tcely:
As far as I can tell, this option only caused a lot of JSON output to be logged.


meeb:
You might be right, thanks!

PR #591, opened 2024-12-20
Handle raise_no_formats exception

### LINKED ISSUE #584: Members Only Videos Continuously Failing
Over the weekend one of the YouTube channels I follow with TubeSync added ~460 member only videos and it's brought my TubeSync installation to a stand-still.

Each of the videos is erroring in the `sync.tasks.download_media_metadata` task then rescheduling with a delay. Due to the number of videos this took hours to get through so that any of the other channels I follow had a chance to be processed. I was surprised to that the reschedule delay got reset the next day when the channel was re-processed, and I was back to everything being jammed up.

I've update the `priority` on the `background_task` records for these failing jobs so that the other channels can be processed so far. Short term what's the best way to stop these videos being continuously processed? Can a mark them as skipped in the `sync_media` table and delete the `background_task` records, or is there a better way?

Longer term there doesn't seem to be any indication in the YouTube api video data to show that these videos require membership, but the following error is being thrown 

```
[tubesync/ERROR] ERROR: [youtube] t4Ukm9YDADQ: This video is available to this channel's members on level: LTT Member Plus (or any higher level). Join this channel to get access to members-only content and other exclusive perks.
Rescheduling Downloading metadata for "b83d27b2-b7f6-42ba-b615-5ca14ac21549"
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/dist-packages/background_task/tasks.py", line 43, in bg_runner
    func(*args, **kwargs)
  File "/app/sync/tasks.py", line 291, in download_media_metadata
    metadata = media.index_metadata()
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/sync/models.py", line 1494, in index_metadata
    return indexer(self.url)
           ^^^^^^^^^^^^^^^^^
  File "/app/sync/youtube.py", line 89, in get_media_info
    raise YouTubeError(f'Failed to extract_info for "{url}": No metadata was '
sync.youtube.YouTubeError: Failed to extract_info for "https://www.youtube.com/watch?v=t4Ukm9YDADQ": No metadata was returned by youtube-dl, check for error messages in the logs above. This task will be retried later with an exponential backoff.
Rescheduling task Downloading metadata for "b83d27b2-b7f6-42ba-b615-5ca14ac21549" for 0:04:21 later at 2024-12-15 19:29:19.699731+00:00
```

I don't know if there's enough information in the error raised to trap this reliably and skip continually retrying to fetch the metadata, bearing in mind someone who is a channel member and has setup their cookies correctly will want to be able to download the videos? I'm guessing they will just download without any error.

Here are some video id's for investigation/testing, let me know if you need any more information.

`t4Ukm9YDADQ`, `T-2eGKOdYz4`, `suL9N97w_Dw`, `sQ1447P7G24`, `sQWuwWAyg08`, `SDP8ZwHo5-s`

tcely:
Catching this and checking the message may be the best way to skip members only videos.

~My current understanding of the situation is that this won't work.~ 🙁 The documentation was inaccurate, so setting `ignoreerrors` to `False` caused the exception to be raised. 👍

I can get a better result, by setting `ignore_no_formats_error` to `True` when fetching the metadata.

Actually using that is likely to have unintended consequences in the future. I'd much rather find a way to only accept no formats for members only videos that can't be downloaded.

The next idea is to ignore the error, then parse the results to check for `"availability": "subscriber_only",` before accepting them as useful.

Fixes #584

### DIFF
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -1499,7 +1499,16 @@ class Media(models.Model):
         if not callable(indexer):
             raise Exception(f'Media with source type f"{self.source.source_type}" '
                             f'has no indexer')
-        return indexer(self.url)
+        response = indexer(self.url)
+        no_formats_available = (
+            not response or
+            "formats" not in response.keys() or
+            0 == len(response["formats"])
+        )
+        if no_formats_available:
+            self.can_download = False
+            self.skip = True
+        return response
 
     def calculate_episode_number(self):
         if self.source.source_type == Source.SOURCE_TYPE_YOUTUBE_PLAYLIST:
--- a/tubesync/sync/youtube.py
+++ b/tubesync/sync/youtube.py
@@ -5,7 +5,6 @@
 
 
 import os
-import json
 from pathlib import Path
 from django.conf import settings
 from copy import copy
@@ -85,16 +84,14 @@ def _subscriber_only(msg='', response=None):
             return True
     else:
         # ignore msg entirely
-        try:
-            data = json.loads(str(response))
-        except (TypeError, ValueError, AttributeError):
-            return False
+        if not isinstance(response, dict):
+            raise TypeError(f'response must be a dict, got "{type(response)}" instead')
 
-        if 'availability' not in data.keys():
+        if 'availability' not in response.keys():
             return False
 
         # check for the specific expected value
-        return 'subscriber_only' == data.get('availability')
+        return 'subscriber_only' == response.get('availability')
     return False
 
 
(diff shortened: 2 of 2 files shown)


meeb:
Thanks! Is checking the `availability` field from the API call for a particular string the only way to detect member-only content? I've not had time to look into this myself but it would be preferable if there was a defined flag.

tcely:
I am not aware of a better option. I forgot where I saw it, but availability should have a known set of options.

meeb:
Fair enough, it's in a sensible location if it needs updating in the future. Thanks!

PR #593, opened 2024-12-21
Handle download_media_metadata task failures

RichardHyde:
The following changes handle failures of the `download_media_metadata` tasks after the set number of attempts caused by trying to get the metadata for a YouTube Member-Only video. See issue #584 

In the `media_post_save` signal handler a new `download_media_metadata` task is only added if the media isn't skipped and there isn't already an active `download_media_metadata` task for the media. This get's around the issue of the tasks attempts count being reset when the old task is removed and a new task is created. I added a new `get_media_metadata_task` to `tasks.py` to check for the existing task.

In the `tasks_task_failed` signal handler the media records skip flag is set to true when the failing task is a `download_media_metadata` task. This check could be removed so that the skip flag is set for any media based task, but I didn't want to unintentionally break anything-else incase the Media tasks failures are handled elsewhere that I haven't come across.

I added `download_media_metadata` to the map in `map_task_to_instance` function which has had the side-effect of showing these errors on the tasks page, which may need to be excluded as there's no reason in the error message as to why the tasks are failing, and the videos can't be manually skipped as the media doesn't have any metadata yet so they don't show up in the media page.


tcely (approved):
Looks good to me. Thanks for your contribution. 👍

meeb:
Yep looks good, thanks!

PR #594, opened 2024-12-21
Ignore case when checking the filter

RichardHyde:
Make the checks against the sources filter text case insensitive.

### DIFF
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -597,7 +597,7 @@ class Source(models.Model):
     def is_regex_match(self, media_item_title):
         if not self.filter_text:
             return True
-        return bool(re.search(self.filter_text, media_item_title))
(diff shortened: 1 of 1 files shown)


tcely (commented):
Making the matching in filtering case-insensitive actually closes off options. It's already possible to match both cases.

RichardHyde:
OK, it caught me out not realising that the filter was case-sensitive, so I thought it was worth a try 😀

tcely:
I predict a new issue is forthcoming about how this breaks someone's filtering that is currently depending on case differences. 😉

meeb:
Huh, I didn't mean to merge this for now. I had all the PRs open in different tabs and specifically didn't merge this. Maybe I've accidentality found a minor CSRF issue with GitHub... I'll undo it for now with a comment.

PR #595, opened 2024-12-21
Do not try to skip already skipped media

tcely:
Logically these functions can only mark media instances as skipped, so running them for media instances that are already marked that way is a waste of resources.


RichardHyde:
The check on `manual_skip` isn't necessary as `filter_media` is only called from the `media_post_save` signal handler, which bails earlyl if `manual_skip` is true

tcely:
That's good to know. Thanks.

meeb:
Seems like a reasonable optimisation, thanks!

PR #597, opened 2024-12-22
Update ci.yaml

tcely:
* Use the current major version tag for actions.
* Add current python versions.
* ~Remove Python `3.7` as it isn't being setup.~
* Use `ubuntu-22.04` to address the `3.7` failure.
* Don't fail fast, so that one old version won't prevent testing other versions.

The most recent `ci` failure from `main`:
```
Run actions/setup-python@v2
Version 3.[7](https://github.com/meeb/tubesync/actions/runs/12452525319/job/34761584494#step:3:8) was not found in the local cache
Error: Version 3.7 with arch x64 not found
The list of all available versions can be found here: https://raw.githubusercontent.com/actions/python-versions/main/versions-manifest.json
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,11 +11,11 @@ on:
 
 jobs:
   test:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


tcely:
@meeb I highly recommend reviewing this before anything else is merged. The tests are failing because of a change in the version of `ubuntu-latest`.

meeb:
I run the tests locally but sure I'll review this next, thanks!

PR #598, opened 2024-12-22
Remove files in post_delete when Media is deleted

### LINKED ISSUE #206: Not deleting media older than "days to keep"

tcely:
I believe this solves the glob failures that meant TubeSync never deleted things from the disk for me.

My format string includes `_[{key}]_` for a plugin that I use with Jellyfin, so that was causing glob/fnmatch to never work on my files.

Fixes #206


meeb:
Thanks!

PR #599, opened 2024-12-22
Fix paths for when filename includes sub-directories

### LINKED ISSUE #294: NFO/jpg not saving to same folder as content/metadata.
Hello.
Odd issue. When I set the media format to nest the files inside of a folder (I use the year), it doesn't place the extra nfo and jpg files in the same folder as the video and metadata files; it places them in the root folder.
The media format I'm using: Season {yyyy}/{title} {mm}{dd} [{key}].{ext}

Screenshot of directories:
image

image

tcely:
Fixes #294


meeb:
Thanks! Looks fine to me.

PR #600, opened 2024-12-23
Do not log download progress lines by default

tcely:
When you set debugging, the default progress output can be viewed and/or logged, depending on the other options.


meeb:
Why would you want to disable the progress logging for downloads?

tcely:
> Why would you want to disable the progress logging for downloads?

It's redundant to have both the download progress hook and the raw terminal status updates in the logs.

meeb:
Ah right, gotcha, didn't think it was duplicated output. Thanks!

PR #601, opened 2024-12-23
Only log new media

tcely:
Channels with thousands of videos, that won't be downloaded, create large blocks in the logs without this.

### DIFF
--- a/tubesync/sync/tasks.py
+++ b/tubesync/sync/tasks.py
@@ -193,7 +193,12 @@ def index_source_task(source_id):
             media.save()
             log.debug(f'Indexed media: {source} / {media}')
             # log the new media instances
-            if media.created >= source.last_crawl:
+            new_media_instance = (
+                media.created and
+                source.last_crawl and
+                media.created >= source.last_crawl
+            )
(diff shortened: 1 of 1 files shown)


meeb:
`source.last_crawl` can be `None` so this probably needs an `if source.last_crawl ...` check as well otherwise it could raise a TypeError in some situations.

tcely:
> `source.last_crawl` can be `None` so this probably needs an `if source.last_crawl ...` check as well otherwise it could raise a TypeError in some situations.

How? It was just set a few lines up.

meeb:
Sure, in this particular function it almost never will be `None`, the only chance being an extremely unlikely race condition, it can be cleared in the admin though and there was a script that bulk reset `last_crawl` to `None` at one point if I remember. Basically I just meant accounting for it as a possibility if the database field can be `None` is probably reasonable.

tcely:
> Basically I just meant accounting for it as a possibility if the database field can be `None` is probably reasonable.

As a general practice, sure. Checking here won't hurt anything.

meeb:
Thanks!

PR #603, opened 2024-12-24
media_file.name uses a relative path

tcely:
Has this worked before?

Nothing I've read has said an absolute path will work when assigned to `media_file.name` but I haven't tried it.

### DIFF
--- a/tubesync/sync/management/commands/import-existing-media.py
+++ b/tubesync/sync/management/commands/import-existing-media.py
@@ -49,7 +49,7 @@ class Command(BaseCommand):
             for filepath, item in filemap.items():
                 log.info(f'Matched on-disk file: {filepath} '
                          f'to media item: {item.source} / {item}')
-                item.media_file.name = filepath
+                item.media_file.name = str(Path(filepath).relative_to(item.media_file.storage.location))
                 item.downloaded = True
                 item.save()
         log.info('Done')
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -644,7 +644,7 @@ def get_media_thumb_path(instance, filename):
 
 
 def get_media_file_path(instance, filename):
-    return instance.filepath
+    return instance.filepath.relative_to(instance.media_file.storage.location)
 
 
 class Media(models.Model):


RichardHyde:
I'm not overly familiar with Djangos filefield type, but from what I've read (https://zerotobyte.com/django-filefield-everything-you-need-to-know/) the path part of the file is set in the `upload_to` property of the field and is stripped/appended to the file name by Django when the record is saved/read from the database. So an absolute path would correct.

https://github.com/meeb/tubesync/blob/904c57f603796679c3cadf14a65b2351e77cc92f/tubesync/sync/models.py#L821-L829

If anything I'd have expected that `upload_to` should be set to the `source.directory_path` rather than `filepath`, which is what`get_media_file_path` returns and includes the file name. That may explain why it's working when importing missing files as the Medias file name would be empty and `filepath` would contain just the `source.directory`, but I wouldn't go fixing that until someone  (@meeb) with more knowledge of the workings of filefield has reviewed it.

tcely:
My question was about `.name` which is how we store new values to the database.

As for `upload_to` it's a callable, not a string, so it's expecting an absolute path to a filename.

https://github.com/django/django/blob/3591e1c1acbd7c13174275367c3fdf012cb0413b/django/db/models/fields/files.py#L317

https://github.com/django/django/blob/3591e1c1acbd7c13174275367c3fdf012cb0413b/django/db/models/fields/files.py#L321

https://github.com/django/django/blob/stable/3.2.x/django/core/files/utils.py#L12

The `storage.location` determines where the relative paths should begin.

https://github.com/meeb/tubesync/blob/904c57f603796679c3cadf14a65b2351e77cc92f/tubesync/sync/models.py#L28

meeb:
This import script used absolute paths because people used this tool to effectively just mark media items as downloaded by unique key. These media items usually weren't in the same `/download` or whatever root `FileSystemStorage` defined structure used by tubesync. For example they would "import" media from `/some/other/mapped/dir/downloads` (as an additional volume mount usually) so the file path stored in tubesync would be the absolute path to `/some/other/mapped/dir/downloads/whatever-media_XXXXXKEY.mkv`. This saved in Django fine and marked the item as "downloaded". This was a hack where using an absolute path overwrites the `upload_to` prefix entirely.

This does generate warnings in some situations where stored media item records in the DB are outside of the `FileSystemStorage` location, however short of allowing arbitrary creation of `FileSystemStorage`s or dropping security to set `FileSystemStorage` to `/` and allow uploads to anywhere on the file system I left it at some occasional warnings and having to create a migration of you change `DOWNLOAD_ROOT` as the side-effects.

Making this relative to the download root would mean it would only work if people moved their media into the `/download` (or wherever `DOWNLOAD_ROOT` is set to) structure first. Plus I think `Path.relative_to(...)` raises a `ValueError` if the path is not a sub-path of the other?

tcely:
> Plus I think Path.relative_to(...) raises a ValueError if the path is not a sub-path of the other?

It does.

I'll close this. I still think we should avoid the full path used in `upload_to` though.

meeb:
> > Plus I think Path.relative_to(...) raises a ValueError if the path is not a sub-path of the other?
> 
> It does.
> 
> I'll close this. I still think we should avoid the full path used in `upload_to` though.

Feel free to submit a PR for that on its own, it likely would be neater.

PR #604, opened 2024-12-24
Use title from DB

tcely:
We stored a stripped title string in the database, just use that instead of `.name` and calling `.strip()` again.


meeb:
Yep fine with me, thanks!

PR #610, opened 2025-01-02
Match another variation of the message

tcely:
```
{key}: This video is available to this channel's members on level: Level 2: Contributor (or any higher level). Join this YouTube channel from your computer or Android app.
```


meeb:
Thanks!

Issue #609, opened 2024-12-27
NFO file not created


henrikmaurin:
Since about a week, NFO files are no longer created for my downloaded videos.
I have this option ticked for all my channels.

I can see the info embedded in the mkv-file, so there is not a problem with getting the description
For the videos I also save an image for, the image is still created.

EDIT
I also tried ticking the export json file, which works. So it seems to be only the export NFO that has been broken

henrikmaurin:
LOGS:
```
TubeSync  | [MoveFiles] Moving file "/downloads/cache/Linus Tech Tips.(2024-12-26) - I Built The Ultimate Hidden Gaming Room.mkv" to "/downloads/video/TubeSync/Linus Tech Tips/Linus Tech Tips.(2024-12-26) - I Built The Ultimate Hidden Gaming Room.mkv"
TubeSync  | 2024-12-28 13:07:57,191 [tubesync/INFO] Successfully downloaded media: tLhfuc_KCpM (UUID: df313ac9-0ed9-4c2e-b06c-360c47c149e9) to: "/downloads/video/TubeSync/Linus Tech Tips/Linus Tech Tips.(2024-12-26) - I Built The Ultimate Hidden Gaming Room.mkv"
TubeSync  | 2024-12-28 13:07:57,308 [tubesync/INFO] Copying media thumbnail from: /config/media/thumbs/df/df313ac9-0ed9-4c2e-b06c-360c47c149e9.jpg to: /downloads/video/TubeSync/Linus Tech Tips/Linus Tech Tips.(2024-12-26) - I Built The Ultimate Hidden Gaming Room.jpg
TubeSync  | 2024-12-28 13:07:57,348 [tubesync/INFO] Writing media NFO file to: /downloads/video/TubeSync/Linus Tech Tips/Linus Tech Tips.(2024-12-26) - I Built The Ultimate Hidden Gaming Room.nfo
TubeSync  | Rescheduling Downloading media for "I Built The Ultimate Hidden Gaming Room"
TubeSync  | Traceback (most recent call last):
TubeSync  |   File "/usr/local/lib/python3.11/dist-packages/background_task/tasks.py", line 43, in bg_runner
TubeSync  |     func(*args, **kwargs)
TubeSync  |   File "/app/sync/tasks.py", line 449, in download_media
TubeSync  |     write_text_file(media.nfopath, media.nfoxml)
TubeSync  |   File "/app/sync/utils.py", line 129, in write_text_file
[...]

PR #611, opened 2025-01-06
Don't chmod a+r when it already has those permissions

tcely:
This should reduce unnecessary errors like those seen in #609.

As proposed in https://github.com/meeb/tubesync/issues/609#issuecomment-2573270228 this also logs a warning on `PermissionError` exception instead of letting that reschedule the task.

### DIFF
--- a/tubesync/sync/tasks.py
+++ b/tubesync/sync/tasks.py
@@ -446,7 +446,11 @@ def download_media(media_id):
         # If selected, write an NFO file
         if media.source.write_nfo:
             log.info(f'Writing media NFO file to: {media.nfopath}')
-            write_text_file(media.nfopath, media.nfoxml)
+            try:
+                write_text_file(media.nfopath, media.nfoxml)
+            except PermissionError as e:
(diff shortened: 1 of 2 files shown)


ZeblodS:
Is it possible to not raise an error if the permissions are not set to precisely 444?

For instance if the folder has forced rights writes/execute (example 775) in that folder that cannot be changed with a chmod.

tcely:
> Is it possible to not raise an error if the permissions are not set to precisely 444? Maybe just raise a warning.
> 
> For instance if the folder has forced rights writes/execute (example 775) in that folder that cannot be changed with a chmod.

Yes, that's handled by the next commit.

ZeblodS:
> > Is it possible to not raise an error if the permissions are not set to precisely 444? Maybe just raise a warning.
> 
> Yes, that's handled by the next commit.
> 
> > For instance if the folder has forced rights writes/execute (example 775) in that folder that cannot be changed with a chmod.
> 
> In the `755` example, the first commit would never try the `chmod` because `755` contains `444` already. (7=4+3, 5=4+1)

Thanks for the explanation.

meeb:
Looks good, thanks!

PR #614, opened 2025-01-08
Make better use of CPU cache with nginx

tcely:


tcely:
That depends entirely on how many thumbnails are being loaded on the media page. A large enough set of static files being read as fast as possible has spiked this container a few times for me.

PR #615, opened 2025-01-08
Test and log ffmpeg version output earlier

tcely:
Running the ffmpeg in an earlier (hopefully cached) layer should clean up the logs a bit.

On a related note, shadowing the environment variable was causing some confusing log output, so stop doing that as well.


meeb:
Thanks!

Issue #463, opened 2024-02-02
Add playlist position media variable


Funtime60:
Would it be possible to add a media variable such as {Playlist_Number} so external tools could better organize my playlist? I'm trying to convert my old batch scripts I used to run manually to an automatic system and this would be really helpful. Thanks.

meeb:
Unfortunately it's not easily possible. This is discussed at length in some other now closed tickets. There's some effort to have a general arbitrary incremental number available as a variable, but this isn't specifically the index of the video on a playlist.

Makhuta:
Would it be possible to use the position in the playlist for the variable? Since it is already used in for example making the .nfo. Also this is how I understood the question of @Funtime60 , if I am wrong sorry for intervening but the resolve I described I would really appreciate

Funtime60:
I'm not sure what you're suggesting? But also I don't think I ended up using this tool. In the end, I wrote my own tool to manage yt-dlp and do what my scripts used to.

meeb:
The .nfo file "position" counter is basically just an arbitrary counter when the .nfo file is generated, it's not useful for a variable used for filenames which are written incrementally, the counter would repeat itself.

Makhuta:
It wouldn't be "counter" since when you are downloading the playlist from YT the yt-dl (usually) provides you with video position in playlist.
Let's take the playlist in the [example](https://www.youtube.com/playlist?list=PL590L5WQmH8dpP0RyH5pCfIaDEdt9nk7r) when adding source from playlist

![image](https://github.com/user-attachments/assets/ad3ad7fe-e92e-4792-a7fc-a806d5562855)

each video from the playlist has it position in playlist which is most likely used in .nfo episode number (probably also in JSON haven't really tried)

```nfo
<?xml version='1.0' encoding='utf8'?>
<episodedetails>
  <title>První výzkumy DINOSAURŮ!  [JurskýPark] #3</title>
  <showtitle>Gejmr &amp; Jirka - Jurský park </showtitle>
  <season>1</season>
  <episode>3</episode>
  <ratings>
    <rating name="youtube" max="5" default="True">
      <value>None</value>
      <votes>24778</votes>
    </rating>
  </ratings>
  <plot>Instagram: https://www.instagram.com/GEJMR
Facebook: http://www.facebook.com/GEJMRcz
Twitter: https://twitter.com/GEJMRcz
WEB: http://www.GEJMR.eu/

Jirka: https://www.youtube.com/user/jirkakral

[...]

PR #617, opened 2025-01-11
Feature

Makhuta:
## Added feature requested in #463
- added function get_episode_str() that is basically calculating number and returning it as string with zero padding (episode 1 will be "01" etc.)
- now "Media format" can use <code>video_order</code> as episode number in file name
    - also added the new variable to "Available media name variables"

### DIFF
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -589,7 +589,7 @@ class Source(models.Model):
             'key': 'SoMeUnIqUiD',
             'format': '-'.join(fmt),
             'playlist_title': 'Some Playlist Title',
-            'video_order': '1',
+            'video_order': '01',
             'ext': self.extension,
             'resolution': self.source_resolution if self.source_resolution else '',
             'height': '720' if self.source_resolution else '',
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -1527,10 +1527,13 @@ class Media(models.Model):
 
     def get_episode_str(self, use_padding=False):
         episode_number = self.calculate_episode_number()
+        if not episode_number:
+            return ''
+
         if use_padding:
-            return f'{episode_number:02}' if episode_number else ''
+            return f'{episode_number:02}'
         
-        return str(episode_number) if episode_number else ''
+        return str(episode_number)
 
 
 class MediaServer(models.Model):
--- a/tubesync/sync/models.py
+++ b/tubesync/sync/models.py
@@ -589,6 +589,7 @@ class Source(models.Model):
             'key': 'SoMeUnIqUiD',
             'format': '-'.join(fmt),
             'playlist_title': 'Some Playlist Title',
+            'video_order': '1',
             'ext': self.extension,
             'resolution': self.source_resolution if self.source_resolution else '',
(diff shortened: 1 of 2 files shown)


tubesync/sync/models.py
```diff
@@ -1373,8 +1375,7 @@ def nfoxml(self):
         nfo.append(season)
         # episode = number of video in the year
         episode = nfo.makeelement('episode', {})
-        episode_number = self.calculate_episode_number()
-        episode.text = str(episode_number) if episode_number else ''
+        episode.text = self.get_episode_str()
```
tcely, line 1378: How does this change interact with media servers?

Makhuta: It was just to prevent writing duplicated code, because the same thing I used for the "Media format" so I put it as function so it can be used on both places. It could be done differently but I thought this was more elegant

tcely: I think you misunderstood.

Plex, or Emby, or Jellyfin are consuming that `.nfo` file. Your change may affect how those media servers sort or parse the media items.

Makhuta: I understand and the only difference is that instead of "generating" the episode.text with this:
```python
episode_number = self.calculate_episode_number()
episode.text = str(episode_number) if episode_number else ''
```
it calls for function get_episode_str() that is like this:
```python
episode_number = self.calculate_episode_number()
return f'{episode_number:02}' if episode_number else ''
```
which is basically the same thing with only difference that the number will have "0" padding like this 1 -> "01"
the reason is that I don't have to write the exact same thing for the 'video_order' as you've got for "episode.text" so I made function and called it in both places. It will produce the exact same thing (minus the padding "0" but I could change it if you want it). I was fully aware of the Plex or Emby requirements, thats the reason I also try to push this change.

As I said if the padding "0" would be an issue for you let me know I will promptly fix it.

tcely: My understanding was this new format was to be used in file names, so not changing the `.nfo` file contents, at all, would be my preference.

I know prefixing a zero on a number seems like a small change that should just work, but I've seen parsing be finicky about similar things before.

Makhuta: Alright I will make it "togglable" and by default it will use the original format

Makhuta: Now it should be exactly the same (for the .nfo file)

tubesync/sync/models.py
```diff
@@ -1524,6 +1525,13 @@ def calculate_episode_number(self):
                 return position_counter
             position_counter += 1
 
+    def get_episode_str(self, use_padding=False):
+        episode_number = self.calculate_episode_number()
```
tcely, line 1529:
```
if not episode_number:
    return ''
```

Checking for your shared else case first will simplify all your later code in this function.

Makhuta: Oh god you are right 😂 I haven't even considered that mainly because I wanted to use the original logic. Since I've got it opened let me just do it

Makhuta: Should be done

tcely (approved):
Looks good to me.

meeb:
Thanks!

PR #620, opened 2025-01-12
Dockerfile health check without the shell

tcely:
Use the form of health checking that doesn't involve an extra shell on every check.


meeb:
Looks fine to me, thanks.

meeb:
Thanks!

PR #622, opened 2025-01-15
Add environment variables for container

tcely:
- TUBESYNC_SHRINK_NEW
- TUBESYNC_SHRINK_OLD

These must be set to the word 'True' (case insensitive) to enable the setting.

These settings were added in #612 and control the filtering of newly retrieved metadata and/or metadata loaded from the database table for media instances.

Additional testers are welcome to try out the metadata filtering code and report back on their experiences.


meeb:
Thanks!

PR #626, opened 2025-01-16
Don't write 'None' in default rating

tcely:
It's better to not have a rating than to create parsing problems.

An example of what is avoided:

```
<ratings>
  <rating name="youtube" max="5" default="True">
    <value>None</value>
    <votes>16781</votes>
  </rating>
</ratings>
```


meeb:
Ah, good catch, thanks!

PR #627, opened 2025-01-16
Don't write zero into MPAA in .nfo

tcely:
It's a bit odd that `mpaa` is used in a `.nfo` that also uses TV series and episode features, but whatever.


meeb:
Thanks!

PR #628, opened 2025-01-16
Create upgrade_yt-dlp.sh

tcely:
This is knowledge that I don't want to remember every time I want to try the latest version.


meeb:
I generally just use "pipenv update" to update all the local libs for testing. Do you prefer using pip directly?

tcely:
Does pipenv work from inside the container?

meeb:
I was just generally manually reinstalling pipenv in the container to test with but if this works fine happy to merge it if it makes your life easier. Could you stuff an echo'd note at the top of it that says something like "this is a tool to assist development and should not be used to generally update yt-dlp, if you have issues please report them on GitHub" or similar? It's not a great idea if people found this then scripted it to update manually making containers dynamic and hard for us to debug.

tcely:
> Could you stuff an echo'd note at the top of it that says something like "this is a tool to assist development and should not be used to generally update yt-dlp, if you have issues please report them on GitHub" or similar? It's not a great idea if people found this then scripted it to update manually making containers dynamic and hard for us to debug.

I'm thinking of checking for the `TUBESYNC_DEBUG` environment variable, before doing anything in this script.

I'll add a function to print a message to standard error too.

I can certainly see how people may do things that won't help us in the future.

tcely:
Alright. This works with both versions of `pip` available in the image.

If you're good with the message, it can be merged.

meeb:
Yep thanks!

PR #630, opened 2025-01-17
Cache for pipenv & apt

tcely:
- Consolidate apk & apt commands
- Consolidate ENV layers
- Consolidate RUN layers
- Remove unused variables
- Remove packages that are no longer needed
- Bind mount /app/Pipfile

`/cache` is now a `tmpfs` mount that has a `cache` mount on top for `pipenv` to use.

`apt` has `cache` mounts in the standard places:
- /var/lib/apt
- /var/cache/apt

This should significantly speed up the `arm64` builds because two wheels were built on that architecture versus only one for `amd64` builds.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,11 +11,14 @@ ARG SHA256_S6_ARM64="8b22a2eaca4bf0b27a43d36e65c89d2701738f628d1abd0cea5569619f6
 ARG SHA256_S6_NOARCH="6dbcde158a3e78b9bb141d7bcb5ccb421e563523babbe2c64470e76f4fd02dae"
 
 ARG ALPINE_VERSION="latest"
+ARG DEBIAN_VERSION="bookworm-slim"
+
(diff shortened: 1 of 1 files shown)


meeb:
Thanks! Looks fine to me.

tcely:
```
#74 132.0   Building wheel for libsass (setup.py): started
#74 1373.0   Building wheel for libsass (setup.py): finished with status 'done'
#74 1373.0 
#74 1373.1   Created wheel for libsass: filename=libsass-0.23.0-cp311-abi3-linux_aarch64.whl size=12880501 sha256=2313e5a93a91af40a4f318c6ab025e4502a3201fa7fe9352ae6ff99cb1c10f47
#74 1373.1 
#74 1373.1   Stored in directory: /cache/pipenv/wheels/a1/a1/53/148dc44ce1a0cfaee1213108605bea345e84e35cf7cc58ee21
```

That's about 21 minutes to build the `libsass` `.whl` (12.29 MiB) on `arm64` that hopefully won't be in the next build of this stage.

PR #640, opened 2025-01-23
Automated channel_id extraction

tcely:
Yay! No more manual searching for channel IDs.

It fetches the channel ID with `yt-dlp` and switches the type from channel to channel-id for the pre-filled add source page.

This is basically as described in https://github.com/meeb/tubesync/pull/306#issuecomment-1384966047

Closes #306

### DIFF
--- a/tubesync/sync/views.py
+++ b/tubesync/sync/views.py
@@ -297,11 +297,11 @@ class ValidateSourceView(FormView):
             '@' == self.key[0]
         )
         if use_channel_id:
-            source_type = self.source_type
             self.source_type_str = 'youtube-channel-id'
             self.source_type = self.source_types.get(self.source_type_str, None)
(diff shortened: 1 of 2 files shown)


meeb:
Looks fine to me! I had a fork that basically did this two years ago or so, the issue with it was at the time YouTube had sort of meta-pages at some /@username URLs that contained multiple IDs, so a collection of channels and playlists, which made adding them automatically pretty impossible (yt-dlp would either error or return a mix of results). These meta-pages might have been removed though as I've not seen one for a while.

From the code this looks like it just falls back gracefully to the current behaviour?

tcely:
> From the code this looks like it just falls back gracefully to the current behaviour?

That's the intended behavior.

meeb:
Thanks! I'll merge this for now. If there's any complaints it should be pretty obvious :) Nice feature to add, certainly.

PR #644, opened 2025-01-24
Fix filesizeformat prefixes

tcely:
I believe that I have converted all the `filesizeformat` uses.

The new `bytesformat` filter uses the proper IEC prefixes to avoid confusion about which base was used in the calculation.

People who insist on using base-10 with bytes (which we settled on as 2^3 bits) are simply wrong! However, the world is filled with silly and wrong people. It's better to not confuse everyone.


meeb:
Thanks!

PR #648, opened 2025-01-25
Display media we can't download as skipped

tcely:
This helps me easily view videos that have no matching formats because they are only for subscribers.


meeb:
Yep makes sense, thanks.

PR #586, opened 2024-12-17
Rename files after a source format change

tcely:
Fixes #185

PR #649, opened 2025-01-25
Fixed where metadata files are located

ZeroInputCtrl:
While the `Media format` can be changed to include subfolders the location of where metadata does not. 
This PR aims to fix that and gives a fairly easy way via the cli manage script to migrate already downloaded files/metadata to the correct location when changed.
None of the existing tests fail, not sure in these tests which to modify to better test this change or if the existing tests are good enough to prove nothing is broken.

Added a task to move/rename media/metadata to match a changed "Media Format" Added a readme
Changed docker location to be more agnostic
Added a bunch of helper functions for the Makefile to aid in local development in docker (do they really belong here, is there a better place to have these?)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,6 +134,8 @@ dmypy.json
 
 Pipfile.lock
 .vscode/launch.json
+container_config
+container_downloads
 
 # Ignore Jetbrains IDE files
 .idea/
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 python=/usr/bin/env python
-docker=/usr/bin/docker
+docker=/usr/bin/env docker
 name=tubesync
 image=$(name):latest
 
@@ -16,6 +16,8 @@ build:
 	mkdir -p tubesync/static
 	$(python) tubesync/manage.py collectstatic --noinput
 
+migrate:
+	$(python) tubesync/manage.py migrate
 
 clean:
 	rm -rf tubesync/static
@@ -26,16 +28,34 @@ container: clean
 
 
 runcontainer:
-	$(docker) run --rm --name $(name) --env-file dev.env --log-opt max-size=50m -ti -p 4848:4848 $(image)
-
+	@if [ "$(persist)" = "true" ]; then \
+		$(docker) run --rm --name $(name) -v ./container_config:/config -v ./container_downloads:/downloads -v ./tubesync/sync/management/commands:/app/sync/management/commands --env-file dev.env --log-opt max-size=50m -ti -p 4848:4848 -d $(image); \
+		$(MAKE) container_logs;\
+	else \
+		$(docker) run --rm --name $(name) --env-file dev.env --log-opt max-size=50m -ti -p 4848:4848 $(image);\
+	fi
 
 stopcontainer:
-	$(docker) stop $(name)
+	-$(docker) stop $(name)
 
+container_logs:
+	$(docker) logs -f $(name)
+
+container_exec:
+	$(docker) exec -ti $(name) $(exec)
+
+container_manage:
+	$(docker) exec -ti $(name) python3 manage.py $(command)
 
 test: build
 	cd tubesync && $(python) manage.py test --verbosity=2 && cd ..
 
+container_rebuild: stopcontainer container runcontainer
+
+container_reset: stopcontainer container
+	rm -rf container_config
+	rm -rf container_downloads
+	$(MAKE) runcontainer persist=$(persist)
 
 shell:
 	cd tubesync && $(python) manage.py shell
--- a/docs/update-media-paths.md
+++ b/docs/update-media-paths.md
@@ -1,0 +1,40 @@
+# TubeSync
+
+## Advanced usage guide - update media paths from the command line
+
+This command allows you to all media paths. You might want to use
+this if you have renamed the Media Format for a source especially 
+if the folders are changed.
+
+
+## Requirements
+
+You have added some sources and media
+
+## Steps
+
+### 1. Run the update media paths command
+
+Execute the following Django command:
+
+`./manage.py update-media-paths`
+
+When deploying TubeSync inside a container, you can execute this with:
+
+`docker exec -ti tubesync python3 /app/manage.py update-media-paths`
+
+There are options for this command:
+* --source
+  * Only target a single source
+  * Use: `./manage.py update-media-paths --source=source_name_here`
+* --no-rename
+  * Do not rename files, only move them to another directory
+  * No value
+  * Use: `./manage.py update-media-paths --no-rename`
+* --dryrun
+  * Make no changes, only print out
+  * No value
+  * Use:  `./manage.py update-media-paths --dryrun`
+
+This command will log what its doing to the terminal when you run it.
+Highly recommended to do a dryrun and reading the log to understand what will be changed.
--- a/docs/update-media-paths.md
+++ b/docs/update-media-paths.md
@@ -2,14 +2,14 @@
 
 ## Advanced usage guide - update media paths from the command line
 
-This command allows you to all media paths. You might want to use
-this if you have renamed the Media Format for a source especially 
-if the folders are changed.
+This command updates all media paths with what is current in the database.
+This may want to be done when the Media Format has been changed for a source,
+especially if the folder structure has changed.
 
 
 ## Requirements
 
-You have added some sources and media
+At least one source and media is present
 
 ## Steps
 
--- a/tubesync/sync/management/commands/update-media-paths.py
+++ b/tubesync/sync/management/commands/update-media-paths.py
@@ -1,0 +1,82 @@
+import os
+from django.core.management.base import BaseCommand
+from common.logger import log
+from sync.models import Source, Media
+import glob
+import shutil
+
+
+class Command(BaseCommand):
+
+    help = 'Updates media when the media format has changed naming/directory'
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--source",
+            help="Only updates specified source"
+        )
+
+        parser.add_argument(
+            "--no-rename",
+            action="store_true",
+            help= "Do not rename the file (only move)"
+        )
+
+        parser.add_argument(
+            "--dryrun",
+            action="store_true", #treats as boolean
+            help="Don't make any changes, only print outputs"
+        )
+
+    def handle(self, *args, **options):
+        filtered_source = options['source']
+        dry_run = options['dryrun']
+        no_rename = options['no_rename']
+        if dry_run:
+            log.info('Dry run initiated')
+
+        if filtered_source:
+            log.info(f'Searching for source: {filtered_source}')
+            sources = Source.objects.filter(name=filtered_source)
+        else:
+            sources = Source.objects
+
+        for source in sources.order_by('name'):
+            log.info(f'Finding media for source: {source}')
+            for item in Media.objects.filter(source=source, downloaded=True):
+                media_path = os.path.dirname(item.media_file.path) if no_rename else item.media_file.path
+                file_path = os.path.dirname(item.filepath) if no_rename else item.filepath
+                if str(media_path) != str(file_path):
+                    log.info(media_path)
+                    log.info(file_path)
+                    log.info(f'Checking media and metadata to move: {source} / {item}')
+                    media_file_path, media_ext = os.path.splitext(os.path.basename(item.media_file.path))
+                    search_path = os.path.join(source.directory_path, '**', media_file_path)
+                    log.info(search_path)
+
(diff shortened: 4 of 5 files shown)


docs/update-media-paths.md
```diff
@@ -0,0 +1,40 @@
+# TubeSync
+
+## Advanced usage guide - update media paths from the command line
+
+This command allows you to all media paths. You might want to use
```
tcely: Is there a word missing in this sentence?

tcely (changes requested):
The changes to `models.py` break the requirements for `.nfo` and `.jpg` files to share the prefix of the downloaded media.

We can't violate this and still work with media servers correctly.

tcely:
> While the `Media format` can be changed to include subfolders the location of where metadata does not.

The full path to the downloaded media controls where these metadata files are placed.

For source dir: `/downloads/video/TheDeathworlders`
With media format providing: `2025/001/the-deathworlders-chapter-22-warhorse-year-1-part-1_[hWtLyUm5DoQ]_vp9-opus.mkv`

The full path is the combination of both.
We take the prefix:
`/downloads/video/TheDeathworlders/2025/001/the-deathworlders-chapter-22-warhorse-year-1-part-1_[hWtLyUm5DoQ]_vp9-opus`

Then add the extension of the metadata file.

Why are you removing prefix in a loop in your function?

ZeroInputCtrl:
> The changes to `models.py` break the requirements for `.nfo` and `.jpg` files to share the prefix of the downloaded media.
> 
> We can't violate this and still work with media servers correctly.

The intention was to keep everything inline, i think maybe the issue i'm trying to fix has been fixed by now (i can't seem to replicate it against main locally). I wrote this a few months ago but didn't re-check the issue on the main before submitting 🤦 . Which was to make sure that the metadata files stayed with the media always (i was having issues between renaming the Media Format, with media and metadata ending up in different folders).

 1. I would still like to have/keep the script to move files based on the current database state for those of us that absolutely want everything to be in-sync with the current settings and add an option to cleanup empty folders after moving media?. 
 2.  i'll drop the changes to the models.
 3. makefile questions still remain, yay or nay?
 4. new PR or edit this one?

tcely:
> The intention was to keep everything inline, i think maybe the issue i'm trying to fix has been fixed by now (i can't seem to replicate it against main locally).

Yeah, I noticed and fixed a few problems with subdirectories when I started using them in the `media_format` myself. #599 

Also, I have been using my automated renaming changes for a while, but that hasn't been merged yet. #586

Makefile
```diff
@@ -26,16 +28,34 @@ container: clean
 
 
 runcontainer:
-	$(docker) run --rm --name $(name) --env-file dev.env --log-opt max-size=50m -ti -p 4848:4848 $(image)
-
+	@if [ "$(persist)" = "true" ]; then \
+		$(docker) run --rm --name $(name) -v ./container_config:/config -v ./container_downloads:/downloads -v ./tubesync/sync/management/commands:/app/sync/management/commands --env-file dev.env --log-opt max-size=50m -ti -p 4848:4848 -d $(image); \
```
tcely, line 32:
```suggestion
		$(docker) run --rm --name $(name) -v $(name)_config:/config -v $(name)_downloads:/downloads -v ./tubesync/sync/management/commands:/app/sync/management/commands --env-file dev.env --log-opt max-size=50m -ti -p 4848:4848 -d $(image); \
```

Using named volumes is a cleaner solution than bind mounted directories and doesn't require additional `.gitignore` entries.

Makefile
```diff
@@ -26,16 +28,34 @@ container: clean
...
+		$(MAKE) container_logs;\
```
tcely, line 33: Why are you running a logs watcher after the container exits?

Makefile
```diff
@@ -26,16 +28,34 @@ container: clean
...
+	else \
+		$(docker) run --rm --name $(name) --env-file dev.env --log-opt max-size=50m -ti -p 4848:4848 $(image);\
+	fi
 
 stopcontainer:
-	$(docker) stop $(name)
+	-$(docker) stop $(name)
```
tcely, line 39: ~Remind us what the '-' here is supposed to do, please.~

`-` suppresses errors. Read more at: [GNU make manual](https://www.gnu.org/software/make/manual/html_node/Errors.html)

Why is `docker stop` returning an error?

Makefile
```diff
@@ -26,16 +28,34 @@ container: clean
...
+
+container_logs:
+	$(docker) logs -f $(name)
 
+container_exec:
+	$(docker) exec -ti $(name) $(exec)
```
tcely, line 45: Where is the value for `$(exec)` coming from?

Makefile
```diff
@@ -26,16 +28,34 @@ container: clean
...
+
+container_manage:
+	$(docker) exec -ti $(name) python3 manage.py $(command)
```
tcely, line 48: Similarly to `container_exec` above.
Where is the value for `$(command)` coming from?

Makefile
```diff
@@ -26,16 +28,34 @@ container: clean
...
 
 test: build
 	cd tubesync && $(python) manage.py test --verbosity=2 && cd ..
 
+container_rebuild: stopcontainer container runcontainer
+
+container_reset: stopcontainer container
+	rm -rf container_config
+	rm -rf container_downloads
+	$(MAKE) runcontainer persist=$(persist)
```
tcely, line 58: This seems unnecessary.
Does using `$(MAKE)` really not preserve variables?

.gitignore
```diff
@@ -134,6 +134,8 @@ dmypy.json
 
 Pipfile.lock
 .vscode/launch.json
+container_config
+container_downloads
```
tcely, line 138: 👎

tubesync/sync/models.py
```diff
@@ -1287,6 +1287,19 @@ def playlist_title(self):
...
+        if self.downloaded and self.media_file:
+            filename = self.media_file.path
+            for prefix in self.source.directory_path.as_posix().split('/'):
+                if prefix == '':
+                    continue
+                filename = filename.removeprefix(f'/{prefix}')
+            filename = filename.removeprefix('/')
+        else:
+            filename = self.filename
+        return os.path.splitext(filename)[0]
```
tcely, line 1301: The `basename` that was removed here causes the duplication of subdirectories when the full path is constructed.

tubesync/sync/models.py
```diff
@@ -1515,9 +1513,15 @@ def download_media(self):
...
+        # Ensure if media is re-downloaded it's saved to the same location that the database says it is
+        if self.media_file:
+            filepath = self.media_file.path
+        else:
+            filepath = self.filepath
+
         # Download the media with youtube-dl
         download_youtube_media(self.url, format_str, self.source.extension,
-                               str(self.filepath), self.source.write_json,
+                               str(filepath), self.source.write_json,
```
tcely, line 1524: I hate this mangling of the arguments.

tubesync/sync/management/commands/update-media-paths.py
```diff
@@ -0,0 +1,82 @@
...
+        parser.add_argument(
+            "--source",
+            help="Only updates specified source"
+        )
+
+        parser.add_argument(
+            "--no-rename",
+            action="store_true",
+            help= "Do not rename the file (only move)"
+        )
```
tcely, line 23: How does this not cause problems when the `media_format` has changed?

tubesync/sync/management/commands/update-media-paths.py
```diff
@@ -0,0 +1,82 @@
...
+    def handle(self, *args, **options):
+        filtered_source = options['source']
+        dry_run = options['dryrun']
+        no_rename = options['no_rename']
+        if dry_run:
+            log.info('Dry run initiated')
+
+        if filtered_source:
+            log.info(f'Searching for source: {filtered_source}')
+            sources = Source.objects.filter(name=filtered_source)
```
tcely, line 40:
```suggestion
            sources = Source.objects.filter(name__contains=filtered_source)
```

I don't want to remember the full name and pass it to your tool.

Also, when doing this from the command line I tend to use `directory` because that is what I see on the filesystem.

tubesync/sync/management/commands/update-media-paths.py
```diff
@@ -0,0 +1,82 @@
...
+        else:
+            sources = Source.objects
```
tcely, line 42:
```suggestion
            sources = Source.objects.all()
```

You can remove this `else:` case entirely by defining `sources` first, then chaining the optional source filter.

tubesync/sync/management/commands/update-media-paths.py
```diff
@@ -0,0 +1,82 @@
...
+        )
+
+    def handle(self, *args, **options):
+        filtered_source = options['source']
+        dry_run = options['dryrun']
+        no_rename = options['no_rename']
+        if dry_run:
+            log.info('Dry run initiated')
+
+        if filtered_source:
```
tcely, line 38:
```suggestion
        sources = Source.objects.all().order_by('name')
        if filtered_source:
```

tubesync/sync/management/commands/update-media-paths.py
tcely, line 40:
```suggestion
            sources = sources.filter(name=filtered_source)
```

tubesync/sync/management/commands/update-media-paths.py
```diff
@@ -0,0 +1,82 @@
...
+            log.info(f'Searching for source: {filtered_source}')
+            sources = Source.objects.filter(name=filtered_source)
+        else:
+            sources = Source.objects
+
+        for source in sources.order_by('name'):
```
tcely, line 44: Wouldn't adding the `order_by` to `sources` be a better idea?

```suggestion
        for source in sources:
```

tubesync/sync/management/commands/update-media-paths.py
```diff
@@ -0,0 +1,82 @@
...
+                    if not os.path.isdir(os.path.dirname(item.filepath)):
+                        os.makedirs(os.path.dirname(item.filepath))
+
+                    if no_rename:
+                        new_path = os.path.join(file_path, media_file_path)
+                    else:
+                        new_path = os.path.splitext(item.filepath)[0]
+
+                    # Find and move media
+                    for file in glob.glob(f'{search_path}.*', recursive=True):
```
tcely, line 66: This is the same glob failure that I fixed in my pull request. The `search_path` needs quoting for certain characters.

tcely (changes requested):
I didn't read the documentation closely yet.

docs/update-media-paths.md
```diff
@@ -0,0 +1,40 @@
...
+## Advanced usage guide - update media paths from the command line
+
+This command updates all media paths with what is current in the database.
+This may want to be done when the Media Format has been changed for a source,
+especially if the folder structure has changed.
+
+
+## Requirements
+
+At least one source and media is present
```
tcely, line 12: Shouldn't the code for this command just warn the user when there are no sources or no downloaded media items?

docs/update-media-paths.md
```diff
@@ -0,0 +1,40 @@
...
+
+### 1. Run the update media paths command
+
+Execute the following Django command:
+
+`./manage.py update-media-paths`
+
+When deploying TubeSync inside a container, you can execute this with:
+
+`docker exec -ti tubesync python3 /app/manage.py update-media-paths`
```
tcely, line 24:
```suggestion
`docker exec -it tubesync python3 /app/manage.py update-media-paths`
```

I prefer alphabetical ordering when it doesn't make any difference to the software.

docs/update-media-paths.md
```diff
@@ -0,0 +1,40 @@
...
+* --source
+  * Only target a single source
+  * Use: `./manage.py update-media-paths --source=source_name_here`
+* --no-rename
+  * Do not rename files, only move them to another directory
+  * No value
+  * Use: `./manage.py update-media-paths --no-rename`
+* --dryrun
+  * Make no changes, only print out
+  * No value
```
tcely, line 36: Why state "No value" here when you didn't describe the value with `--source` above?

tcely (changes requested):
I've finished reviewing it all.

Splitting this into two new pull requests seems like a good idea.

1. The new command and its documentation
2. The Makefile changes

meeb:
If #586 was merged would a tool like this be necessary?

ZeroInputCtrl:
> If #586 was merged would a tool like this be necessary?
> 
> Edit: and thanks for the PR @ZeroInputCtrl - all contributions are appreciated!

Having it automatic IS a better path to take when the change is made to the config. So no it's not necessary given #586.

ZeroInputCtrl:
👍

PR #653, opened 2025-01-27
Default to valid JSON

tcely:
This prevents useless logged errors when there is no metadata available yet.


meeb:
Yep fair enough.

PR #655, opened 2025-01-27
Use yt_dlp.get_postprocessors()

### LINKED ISSUE #645: Use yt-dlp configuration for SponsorBlock
Instead of adding a post processor, set the configuration options and let `yt-dlp` decide the order of post processing for us.

The UI issue #362 also should be corrected, so that previous selections are displayed reliably.

tcely:
Fixes #645


meeb:
Looks reasonable to me.

I realise the media item instance isn't passed to `download_media(...)` but that string check could match if the video title contains (for example) `av1-` for some reason, however unlikely that might be.

Perhaps the media item instance should be passed to `download_media(...)`? Then you could use the actual codec match from `media.source....`. This could also simplify the args for `download_media(...)` which have grown quite a lot lately. I think in very early drafts `download_media(...)` was called in a worker that couldn't be passed a media instance without an extra lookup but that seems pretty moot now given it's called from where a media instance is already created.

I'm pretty sure `download_media(...)`  is only called in two places so it's not a massive refactor, once in `models.py` as `download_youtube_media(...)` and once in `signals.py` if I remember. Both can just have `self` or `instance` added as the first arg most likely.

tcely:
Refactoring it would be a good idea. Especially when we start asking for `yt-dlp` to select the format for us.

I'd like to get away from passing a format and handling fallbacks, if possible.

meeb:
Thanks!

PR #656, opened 2025-01-27
Set a default nice level

tcely:
Now that SponsorBlock is running, these jobs are using significant CPU resources.

Setting a nice level to let other things run better while `ffmpeg` is running seems wise.


meeb:
Yep seems reasonable.

PR #657, opened 2025-01-29
Fix CommaSepChoiceField

tcely:
Using the separator that was chosen is the biggest fix.


meeb:
Heh, thanks. Good catch.

tcely:
Thanks. Please take care off the migration this has likely caused.

PR #658, opened 2025-01-29
Use TextChoices for SponsorBlock_Category

tcely:


meeb:
Looks fine to me, thanks.

PR #659, opened 2025-01-29
Use a temporary directory per task

tcely:
I am seeing multiple ffmpeg processes that mess each other up.

I do not know why this is happening yet, but for now let them operate in their own temporary directories.


meeb:
That does sound weird. I've never encountered that myself. Given ffmpeg is only spawned by yt-dlp this might be upstream.

tcely:
Sadly, if it were upstream, this solution would not help. It does and I see two tasks listed on the tasks page for the same video key also.

Something is going wrong on our side, I just haven't found how it is happening yet.

meeb:
Yeah if you see two tasks for it then it's definitely local.

PR #663, opened 2025-01-29
Reorganize hooks

### LINKED ISSUE #662: Add hook for post processing
We have one for downloading already, with SponsorBlock working it can take a while.

It's better to log something about how that is going.

```
postprocessor_hooks:  A list of functions that get called on postprocessing
                       progress, with a dictionary with the entries
                       * status: One of "started", "processing", or "finished".
                                 Check this first and ignore unknown values.
                       * postprocessor: Name of the postprocessor
                       * info_dict: The extracted info_dict

                       Progress hooks are guaranteed to be called at least twice
                       (with status "started" and "finished") if the processing is successful.
```

tcely:
This has gotten a bit larger than I figured it would.

Further logging / improvement can happen in a future pull request.

Fixes #662


meeb:
Yep fair enough, prettier logging is nice.

PR #664, opened 2025-01-30
Add `ytopts`

tcely:
These are all optimizations:
- `writethumbnail`: for EmbedThumbnail to use
- `check_formats`: to save requests
- `overwrites`: to keep the first version of the video
- sleep intervals to slow down our tasks only for video downloads


meeb:
Yep looks sensible. Thanks.

PR #667, opened 2025-01-31
Fix log.debug()

tcely:
`common.logger` imports `settings`

Creating a circular dependency causes the failure to load this file from `tubesync/settings.py`.

Without the change to `DEBUG` from this file the log level is `INFO` instead of `DEBUG` so none of those messages were showing.

This was broken by 20df9f4044a3c66f641b2de145efc1b15e0ece56.


meeb:
Thanks!

PR #672, opened 2025-01-31
Adjustments for downloading delays

tcely:
I was much too generous with these delays. There are many more requests affected than I believed.

Downloads sleep between 10 and 90 (or longer when `MAX_RUN_TIME` is above 600) seconds.

Every other request, sleeps for 5 seconds.


meeb:
Yep makes sense.

PR #673, opened 2025-02-01
Configure settings for this test

tcely:
Otherwise, environment variables can cause the `unfiltered` and `filtered` responses to match. This breaks the assumption that the asserts are depending upon.


meeb:
Thanks!

PR #675, opened 2025-02-01
Last attempt to preserve that `log` line

tcely:
Logging really should not be done in settings.

If this doesn't work around both competing problems, then we can log this information elsewhere instead.


meeb:
I just pushed https://github.com/meeb/tubesync/commit/5e1da4e58e80f16f7d5f2f01cfb827bc4bb966cf

PR #685, opened 2025-02-02
Wait for video premiere

### LINKED ISSUE #676: Smarter rescheduling of metadata tasks
```
sync.youtube.YouTubeError: Failed to extract_info for "https://www.youtube.com/watch?v=KEY": ERROR: [youtube] KEY: Premieres in 2 days                          

Rescheduling task Downloading metadata for "UUID" for 1:49:26 later at 2025-02-01 06:54:54.353598+00:00
```

Ideally, there should be a new task scheduled for two days later and TubeSync should not try this video again until after that task runs.

Perhaps, setting `manual_skip` then unsetting that again with the scheduled task.

tcely:
Starting on the smarter scheduling for upcoming videos.

Screenshot_20250202-231240

This quick and messy solution appears to be working. The basic approach is sound.

I would still like to edit the task name from within the task function to display the remaining hours in the task list also.

Fixes #676

### DIFF
--- a/tubesync/sync/tasks.py
+++ b/tubesync/sync/tasks.py
@@ -317,13 +317,16 @@ def download_media_metadata(media_id):
         metadata = media.index_metadata()
     except YouTubeError as e:
         e_str = str(e)
+        log_exception = True
         if ': Premieres in ' in e_str:
             now = timezone.now()
             published_datetime = None
 
-            parts = e_str.rsplit(' ', 2)
-            unit = lambda p: str(p)[-1].lower()
-            number = lambda p: int(str(p)[-2], base=10)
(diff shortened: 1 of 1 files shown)


meeb:
Looks reasonable to me, thanks! I'm never that on board with the string checks in yt-dlp output to trigger special behavior but I realise that's regularly the only way to detect certain errors. Do you want this merged now?

tcely:
> I'm never that on board with the string checks in `yt-dlp` output to trigger special behavior but I realise that's regularly the only way to detect certain errors.

I'm not a fan of that approach either. This is fragile, but breaking only means that we fail to improve the situation, not that the video doesn't download.

> Do you want this merged now?

Yes. Thanks.

PR #686, opened 2025-02-03
Turn off the `text-transform` with a class

tcely:
It turns out using a `div` in a06bd29f88d46eec86de071c3495d49d932f1707 broke the interface.

This is a better way.


tcely:
@meeb Any objections to merging this?

meeb:
Nope, thanks!

PR #690, opened 2025-02-05
Let Tasks expire correctly

### LINKED ISSUE #252: locked tasks from crashed process_tasks process not unlocked
Situation:

- 22 sources, ~9700 media items
- TZ: UTC (matches host tz)
- TUBESYNC_WORKERS: 8
- DATABASE_CONNECTION: postgresql://[...]

Sometimes the `process_tasks` process gets killed/exits (for whatever reason) and then is restarted by s6. However, if the original process had locked tasks in the db, then these do not get unlocked, not on restart and not after they are expired. Should the amount of locked 'dead' tasks be equal to `BACKGROUND_TASK_ASYNC_THREADS`, then the whole task queue is deadlocked until tasks are reset. 

As far as I can tell, this is because of a bug in `django-background-tasks`, related to: https://github.com/arteria/django-background-tasks/issues/239

When using `BACKGROUND_TASK_ASYNC_THREADS` tasks never expire as per the linked issue. So even if there are locked tasks in the db (by processes which do not longer exist), they wont get unlocked. Same for tasks running longer than `MAX_RUN_TIME` (thats what the linked issue is actually about).

Ideas to solve this or work around this:

- fix this in `django-background-tasks` or patch it in the version you are using (since it is unmaintained)
- create an additonal command/script within the container (and run it via s6), which watches the tasks of the taskmanager via `task.locked_by_pid_running()` and unlocks them, should it return false.

tcely:
The `locked` function is using `|` instead of `&`, so we eliminate the `locked_by` filter to let `locked_at` work correctly.

Viewing the tasks page will unlock tasks immediately when the locking process no longer exists.

[Upstream PR](https://github.com/django-background-tasks/django-background-tasks/pull/305)

Fixes #252


meeb:
I'm not opposed to upstream patching but this patch would need to be added to a `/patches/` dir in this repo probably. Having the container built depending on other accounts is a bad idea. Not that I don't generally trust your patches at this point, just as a general principle.

tcely:
That's fine. I'll upload that file later and adjust accordingly.

tcely:
That should be it.

tcely:
@meeb The conflict has been resolved.

meeb:
Thanks! I'd still like to see the diff to know what you've changed from upstream if possible.

tcely:
I linked the file name to the commit.

https://github.com/tcely/django-background-tasks/commit/f9ec681dc4e4d92ca0a78b89450da9dd871e82a7

meeb:
Ah, that one line change is it? Fair enough. Could this just be a monkeypatch rather than overwriting the whole file?

tcely:
It could be, but the `patch` program isn't available in the image.

I didn't see a point to creating a stage to create a file to copy when I could just download the whole file from the branch.

meeb:
I meant just embedding the patch in tubesync directly as a monkeypatch, like:

```
from background_task.settings import app_settings
from background_task.models import TaskManager
from other.imports.as.needed import blah

def _locked_fixed(obj, now):
    max_run_time = app_settings.BACKGROUND_TASK_MAX_RUN_TIME
    qs = obj.get_queryset()
    expires_at = now - timedelta(seconds=max_run_time)
    locked = Q(locked_by__isnull=False) & Q(locked_at__gt=expires_at)
    return qs.filter(locked)

TaskManager.locked = _locked_fixed
```

or whatever (untested hacky example). No `patch` or additional build stage needed.

tcely:
Oh. I wasn't considering anything quite as hackish as that.

It seems like a lot more work for a less clean solution.

tcely:
@meeb I would prefer to merge this.

Changing the behavior during run time is more surprising for people than changing the source.

How we modify the source, we can change or not. Using a full file is easiest for use in the `Dockerfile`, so that's what I went with.

meeb:
OK I'll merge this. I suggested the monkeypatch option as it generally is less likely to break if upstream gets updated, plus overwriting an entire file or a one line change is a little excessive. However, given the upstream seems pretty much dormant it likely doesn't matter much.

PR #691, opened 2025-02-06
Rework download progress hook

tcely:
- Be more resilient about missing keys.
- Log the downloading state the first time through.
- Keep `hook.download_progress` under 100.
- Try to use `total_bytes_estimate` when `total_bytes` is not available.

~For multiple downloads, at the same time, this needs to be improved further.
Each download will need to track status independently.~
The key may need to be adjusted, but the basic support is already present.

Reference:
```
    progress_hooks:    A list of functions that get called on download
                       progress, with a dictionary with the entries
                       * status: One of "downloading", "error", or "finished".
                                 Check this first and ignore unknown values.
                       * info_dict: The extracted info_dict

                       If status is one of "downloading", or "finished", the
                       following properties may also be present:
                       * filename: The final filename (always present)
                       * tmpfilename: The filename we're currently writing to
                       * downloaded_bytes: Bytes on disk
                       * total_bytes: Size of the whole file, None if unknown
                       * total_bytes_estimate: Guess of the eventual file size,
                                               None if unavailable.
                       * elapsed: The number of seconds since download started.
                       * eta: The estimated time in seconds, None if unknown
                       * speed: The download speed in bytes/second, None if
                                unknown
                       * fragment_index: The counter of the currently
                                         downloaded video fragment.
                       * fragment_count: The number of fragments (= individual
                                         files that will be merged)

                       Progress hooks are guaranteed to be called at least once
                       (with status "finished") if the download is successful.
```

### DIFF
--- a/tubesync/sync/hooks.py
+++ b/tubesync/sync/hooks.py
@@ -30,7 +30,6 @@ class PPHookStatus:
 
 def yt_dlp_progress_hook(event):
     hook = progress_hook.get('status', None)
-    filename = os.path.basename(event['filename'])
     if hook is None:
         log.error('yt_dlp_progress_hook: failed to get hook status object')
         return None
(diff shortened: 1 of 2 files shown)


meeb:
Looks fine to me, thanks!

PR #697, opened 2025-02-07
Use `libopus` and speed up VP9 encoding

tcely:
I just saw a 300% increase in encoding speed.

I will never understand why the VP9 defaults are what they are. Good results should not require this many flags.

Patching the [`modify_chapters.py`](https://github.com/tcely/yt-dlp/blob/modify-chapters-codecs/yt_dlp/postprocessor/modify_chapters.py) file allows us to eliminate an extra encoding step. [Upstream PR](https://github.com/yt-dlp/yt-dlp/pull/12317)


meeb:
Looks like a nice enhancement to me, thanks! This works on ARM as well and doesn't affect any other codecs?

tcely:
> This works on ARM as well and doesn't affect any other codecs?

That's a very fair question! It's going to depend on how `ffmpeg` was built for ARM.

If it doesn't, it's easy to override these flags from the settings. I chose the most specific key I could for exactly that reason.

meeb:
Seems to have reasonably comprehensive cover then, thanks! I'll merge it for now and if anyone has some weird ARM build they can let us know later...

PR #700, opened 2025-02-09
Remove subtitles also

### LINKED ISSUE #416: Deleting media does not deletes subtitles
When medias are deleted, such as when removing a source, the subtitles are not removed. 

We should check if the source has enabled subtitles and delete all `{filename}.vtt` files too

tcely:
https://github.com/meeb/tubesync/blob/8d89cc01f961b6440584d5984784e01881dfa900/tubesync/sync/signals.py#L263-L268

I think this was already covered by `all_related_files` at the end, but matching `*.vtt` files earlier won't hurt anything.

Fixes #416


meeb:
Thanks!

PR #701, opened 2025-02-09
Allow `exec_cmd` dictionary to be configured in settings

tcely:
This is a dictionary with keys of `WHEN` and values are a `list` of commands for the `ExecPP` to run.


meeb:
Thanks!

PR #704, opened 2025-02-09
Try to include `timestamp`

tcely:


tubesync/sync/models.py
```diff
@@ -944,7 +949,7 @@ def save(self, force_insert=False, force_update=False, using=None, update_fields
 
     def get_metadata_field(self, field):
         fields = self.METADATA_FIELDS.get(field, {})
-        return fields.get(self.source.source_type, '')
```
tcely, line 947: @meeb Was there a good reason this returned `''` for a `field` that wasn't mapped?

meeb: Just that all metadata values were strings in the first super beta version of tubesync, it was just conformist and still evaluated to False in logic tests.

meeb:
Seems fine, thanks!

PR #710, opened 2025-02-11
Deduplicated `fields`

tcely:


meeb:
I'd rather this was a big explicit list than the magic defined here if I'm honest. Does this save some hassle I'm not seeing?

tcely:
Alright. The manual ordering did still contain duplicated fields.

meeb:
Thanks!

PR #712, opened 2025-02-11
Control `writethumbnail` with `embed_thumbnail`

### LINKED ISSUE #708: *.webp file to every download
Hello,

since 2 or 3 month tubesync adds to ever download (only audio or video) a .webp file with the same name.
why is that? how to disable this files?

this messes withe the libary of the musicplayer.

thanks and greetings

tcely:
Fixes #708


meeb:
Thanks!

Issue #70, opened 2021-02-22
Include log in /config folder


PhuriousGeorge:
enhancement req

To help with troubleshooting, it would be helpful to easily retrieve log data (and differentiate log-level).  I haven't entered the container to see if a log is kept internally, but reviewing via `docker logs tubesync` or within an admin interface ie: Portainer, most of what is seen is output stating what it's collecting to generate the web interface.

meeb:
Yep a reasonable suggestion. The only reason this wasn't done already is I didn't want to have to handle log rotation manually in the container with yet another process, without it the log could get massive pretty quickly.

PhuriousGeorge:
Understood.  I've not developed much in Linux beyond some caveman scripts, but isn't log rotation typically managed by the `logrotate` package and a cron job?

meeb:
Yes, via logrotate on the host machine. However the container can't assume logrotate is set up and configured on the host to manage the log rotations properly (e.g. via Docker on Windows or deploying onto a Kubernetes cluster which have no ability to have logrotate for files running). Without some in-container log rotation management the logs TubeSync would write would have by default (from `docker run...`) potentially no maximum size. The Python logging library supports rotation by day, but not zipping etc. features that logrotate does. It's perfectly possible, just a bit of extra work to implement as it has to work entirely out of the box.

PhuriousGeorge:
Since the /config folder is exposed to the guest, I would have thought logratate within the container should be able to manage it? I'm still trying to figure out Docker, but I've managed this in an lxc container with an attached directory.

meeb:
Yes, logrotate can handle it if configured on the host quite easily. There's no way for logrotate no easily notify the container that the log has been rotated, but that's not too much of an issue.

I just don't want to have to require setting up logrotate on the host for people who may not know how to set it up or configure it, or potentially add a "landmine" option where "enabling logging" can do something like fill up your volume on a k8s deployment with no obvious solution where logrotate may not even be available.

The "docker/container" way to handle this would be to just use docker for logging including just to log container stdout/stderr to the hosts syslog (for example with `--log-driver=syslog`) and then configure syslog to write container logs to a logfile if you want them that way by filtering them with a syslog facility. I can see why a plain text file is useful but for ease of use and universal working out of the container without any additional magic required on the host it'll need some extra log rotation handling work.

Code-Slave:
what ive seen done is set a linecount rotate logs when it hits that count

meeb:
[...]

PR #713, opened 2025-02-11
Logs for `nginx` under `/config/log/nginx`

tcely:
For #70

I found [hat-syslog](https://github.com/hat-open/hat-syslog) which looks like an excellent way to handle logs from `nginx` and potentially anything else that uses a `syslog` server.


meeb:
Doesn't this break HTTP request logging for the container in general? As in, removes it from the standard stdout and writes it to disk? Also unless I've missed it this doesn't have any log rotation by default?

tcely:
@meeb The merge conflict has been resolved.

meeb:
Thanks. I'm on the fence with this one, your patch is fine, it's a philosophical difference. I prefer containers to write their logs, all of them, to stdout. I don't think the web server gets enough traffic for this "spam" to be significant. Is there a core reason to divert some of the logs to disk and potentially bypassing general container log analysis tools?

tcely:
I personally almost never care about the web access logs. They are annoying and I'm the way when I want to follow what is happening.

A wishlist item might be to have a way to send logs from the web server through the common logger and provide a way to turn displaying them on/off while everything is running. That would solve both of our use cases.

tcely:
Replaced by #1472.

PR #727, opened 2025-02-14
Do not turn `Media.skip` off for already skipped media

### LINKED ISSUE #723: Re-downloading deleted videos.
Hey meeb! I updated my container last night to whatever had been committed up to 10PM MST on Feb 13th and when the tasks were run this morning, it triggered a re-download of everything that had been deleted from my library. Plex manages my files as I watch them, so that I don't have Tubesync delete something that I haven't seen. It hasn't been an issue in the past, but today it decided to download 300+ videos that had been deleted from the library by Plex. The previous update to my container was I think on Feb 2nd. Let me how to export the logs you need and I'll get them for you. Thanks.

tcely:
~This probably needs a more nuanced solution.~

Ok. I think this handles the cases we need.

Fixes #723

### DIFF
--- a/tubesync/sync/filtering.py
+++ b/tubesync/sync/filtering.py
@@ -12,7 +12,7 @@ from .overrides.custom_filter import filter_custom
 # Check the filter conditions for instance, return is if the Skip property has changed so we can do other things
 def filter_media(instance: Media):
     # Assume we aren't skipping it, if any of these conditions are true, we skip it
-    skip = False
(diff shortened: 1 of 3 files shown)


meeb:
I've not tested this but it looks OK at first glance. I'm guessing this is quite important to merge though given the other issue reporting media redownloading?

meeb:
Ah, it is, OK I'll just merge it for now. Thanks.

PR #730, opened 2025-02-16
Schedule a rename task per media item instead

tcely:
This is a user experience improvement and should also be more effective as well.

Having a task per download gives the user better insight into the progress of the renaming.

It also allows a new task to remove the queued task because the locked task no longer includes every media item.


meeb:
This ready to be merged?

tcely:
Yes. It is working fine on my test instance.

meeb:
Thanks!

PR #736, opened 2025-02-17
Load `fulltitle` from the `metadata`

tcely:


meeb:
Ah looks like this might require an update to the tests.

tcely:
That looks like wasted effort. I'm going with #740 instead.

PR #747, opened 2025-02-19
Add and use `refresh_formats`

### LINKED ISSUE #531: [Improvement] retry "no machting formats"
After adding the ability to download live stream VODs the problem arose that many of the long streams aren't yet finished encoding by YouTube into the desired format and are skipped because of "no machting formats".
It would be great if those videos would be retried to get download with the same logic as if YT-DLP runs into an error while downloading (maybe up to 5-10x times).

### LINKED ISSUE #647: Fetch metadata again after download task times out
For #188 and other "disappearing formats" problems, it would be great if the metadata was pulled again by the software instead of the user skipping/unskipping media items.

I think we should be careful to preserve the existing metadata during this process because the video might have been deleted, and I want to still have the metadata in those cases.

tcely:
Fixes #531
Fixes #647

### DIFF
--- a/tubesync/sync/filtering.py
+++ b/tubesync/sync/filtering.py
@@ -130,14 +130,7 @@ def filter_max_cap(instance: Media):
         return False
 
     max_cap_age = instance.source.download_cap_date
-    if not max_cap_age:
-        log.debug(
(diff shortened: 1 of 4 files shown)


meeb:
Nice, thanks!

PR #754, opened 2025-02-21
Pin `django-background-tasks`

tcely:
It's `django-background-tasks` that limits the Django version.

PR #765, opened 2025-02-24
Upgrade to 1.2.8 for Django 5 support

tcely:
These are the changes used to build:
`ghcr.io/tcely/tubesync:django-5`

See also:
- #754


meeb:
Awesome, thanks!

PR #766, opened 2025-02-24
Remove Redis

tcely:


meeb:
This might be more sensible to just comment out the run commands and comment out the apt package for now, at some point the worker faff probably needs to be replaced regardless of your excellent hacking on it.

tcely:
Resurrection from `git` isn't terribly difficult.

The run commands were disabled already.
However, the presence of the services and files left me confused as to how the tasks were actually running when I first looked at the source.

tcely:
> Sure, I've created a blank new project on tubesync and added you with write access to it. Is that sufficient?

It appears to be. Thanks!

meeb:
I've set the project to public, it appears already to be linked to the repo.

PR #790, opened 2025-02-28
Add `pagination` for scheduled tasks

### LINKED ISSUE #80: Paginate or sumarise /tasks
The tasks page currently straight lists all the queued tasks. If large sources are added these tasks can number in the multiple thousands and make the `/tasks` page vast. This should either be split into tasks per source summary, or paginated, or both.

tcely:
Fixes #80


meeb:
Nice, thanks!

PR #797, opened 2025-02-28
Keep cache between builds

tcely:


meeb:
Can you walk me through what's going on with wormhole and beaming stuff about to make this work? Reviewing the diffs on a phone is going to be painful otherwise.

tcely:
> Can you walk me through what's going on with wormhole and beaming stuff about to make this work? Reviewing the diffs on a phone is going to be painful otherwise.

Sure thing.

Here's the executive summary:

1. We can read the local runner filesystem to access the restored cache files.
2. We can't write to that bind mount.
3. `magic-wormhole` has default infrastructure, but using that would allow modification of our transfers (because the code is logged and passed around)
4. To avoid interference we run the `magic-wormhole` infrastructure in the runner so traffic never leaves the host. To interfere, the adversary needs to access the runner. GitHub is responsible for making that difficult. 😉
5. The runner needs to accept multiple file transfers, so I have a loop running in the background that moves the files after a receiver exits and starts another receiver.
6. The build has a cached virtual environment with `wormhole` installed per platform.
7. The build sends the `/cache/saved` directory it has to the runner, so the `cache` action will save it in the GitHub cache infrastructure for us.

Currently, the virtual environment is the only platform specific piece because of of some compiled dependencies.

tcely:
It was stupidly complex to figure out. However, after I researched what other people put together to solve this problem of accessing the `cache` mounts from the `buildx` builder, this is the far simpler option.

tcely:
> I'm still not that sold on this idea or how it functions.

Alright. I'm going to keep using it and likely fiddling with the idea for my builds.

You should probably merge #808 to help your build times.

tcely:
Another possibility for solving the long compilation time, but nothing else would be:

1. Add a `redis` service 
2. Point `sccache` at that
3. Store the data from `redis` into the GitHub cache

This solution has less for the docker container to do. However, saving and restoring data for a workflow service might be fairly complicated on the runner side of things.

Issue #777, opened 2025-02-25
Store metadata when importing files


tcely:
As a follow-up from #69, it would be great to import any `.info.json` files that may be sitting next to video files we are importing.

PR #805, opened 2025-03-04
Save and use cached `.info.json` files

tcely:
The idea is to reduce the number of requests sent to YouTube.

Related to:
- #777


tcely:
Incorporated into #905.

PR #807, opened 2025-03-04
No more errors for `None` values set in metadata

tcely:


meeb:
Looks like a reasonable solution to an annoying problem, thanks.

PR #808, opened 2025-03-05
Speed up `arm64` builds

tcely:
This is the dependency that takes so long to compile that it dominated every other part of the build time.

It's a ~12 MiB wheel when we compile from the latest source.


meeb:
Yep building libsass is a pain. Thanks.

tcely:
Using this package was the fallback option I had for if the cache preservation pull request was unworkable.

We can still add this change, but it doesn't matter as much anymore. My latest builds are around 10 minutes with the preserved cache working.

Issue #817, opened 2025-03-08
Indexing is really slow


henrikmaurin:
I don't really know how to debug this.
But my _feeling_ is it has gone worse with the later versions...

Before I had no problems with the updates piling up, but now I have a backlog of about 40 of my 60 channels.
Is the problem in yt-dlp itself?
Can I make the code skip all older items, not reindexing them?

Could there be some problem with the database? Can I rebuild it without losing my data?

I have no problem adding logs myself, but I'm not sure where to start

My system:
Running in docker on Open Media Vault with MariaDB as database:

![Image](https://github.com/user-attachments/assets/493ff223-8124-4f23-805e-8a5fd530ee03)

The system disk is a nvme WD BLACK SN770 1TB
Internet is 1 GB/s

This is from my currently runnin channel update (Cap is set at one month)

`tubeSync  | 2025-03-08 11:03:18,368 [tubesync/DEBUG] Indexed media: OutsideXbox / jL4vp_9VzXA
tubeSync  | 2025-03-08 11:03:46,166 [tubesync/DEBUG] Indexed media: OutsideXbox / D4-B1nnVi50
tubeSync  | 2025-03-08 11:04:13,845 [tubesync/DEBUG] Indexed media: OutsideXbox / 1L6v1XRXHiM
tubeSync  | 2025-03-08 11:04:41,587 [tubesync/DEBUG] Indexed media: OutsideXbox / 1iYC1f_1iAw
tubeSync  | 2025-03-08 11:05:09,211 [tubesync/DEBUG] Indexed media: OutsideXbox / vcjXEwxPHRY
tubeSync  | 2025-03-08 11:05:37,303 [tubesync/DEBUG] Indexed media: OutsideXbox / WCzlh3XJhbU
tubeSync  | 2025-03-08 11:06:05,777 [tubesync/DEBUG] Indexed media: OutsideXbox / Z5Fq666-9AA
tubeSync  | 2025-03-08 11:06:33,462 [tubesync/DEBUG] Indexed media: OutsideXbox / aTM9mI-9xYE
tubeSync  | 2025-03-08 11:07:03,446 [tubesync/DEBUG] Indexed media: OutsideXbox / WbA2KpMx4mA
tubeSync  | 2025-03-08 11:07:35,755 [tubesync/DEBUG] Indexed media: OutsideXbox / gLNkOywMgWA
[...]

PR #818, opened 2025-03-08
Patch `yt-dlp` to use `check_thumbnails`

tcely:
Patch to use `check_thumbnails` instead of `check_formats` to mean test downloading every possible thumbnail URL.

I also opened an [upstream issue](https://github.com/yt-dlp/yt-dlp/issues/12584).

For #817


meeb:
Will do! I hadn't noticed the conflict honestly, when I last looked at this it didn't have a conflict. If you're happy with it I'll wait to see what the upstream maintainers say, we may not need a patch if they fix it.

PR #821, opened 2025-03-09
Better environment variable support for settings

tcely:
`os.getenv` makes no guarantees about the return type for default values.


meeb:
Thanks!

PR #831, opened 2025-03-11
Delete pip.conf

tcely:


meeb:
Thanks! Probably fine, we'll know soon enough if we get a load of issues opened :)

PR #842, opened 2025-03-14
Improve deleting sources

### LINKED ISSUE #825: Can't delete a large source using the web UI
If the delete doesn't complete within the timeout it fails.

tcely:
Fixes #825


meeb:
Thanks!

PR #859, opened 2025-03-18
The workflow should work for forks also

tcely:
This automated `FFMPEG_DATE` and solves my periodic build failures when I update my main branch.


meeb:
I generally don't like non-imdepodent container builds, however given the container already runs `apt update` so base packages can change and `yt-dlp` randomly bumps itself on build this probably isn't adding anything that isn't already in the container.

Thanks!

tcely:
I've put together a pull request, so that the ffmpeg build doesn't change every day.

See:
- #870

Issue #375, opened 2023-05-03
add mark all and delete for the souce media list page


gautamkrishnar:
Adding a delete all and skip media button to the media listing page of the source (https://example.com/media?filter=adw32q-79f6-4d9a-a372-a3a723c38dd0) would greatly improve the UX

PR #861, opened 2025-03-18
Transactions are tricky for deleting sources

### LINKED ISSUE #825: Can't delete a large source using the web UI

### LINKED ISSUE #858: Deleting Source leads to 504
When trying to delete a source, the page is displayed with _504 Gateway Time-out._

The affected source has about 6k of elements where 1/3 has been downloaded and not needed anymore.

tcely:
See also:
- #375 

Fixes #825
Fixes #858


tcely:
@meeb 

Ok. I've played with this enough to know none of my ideas will work.

I need a better idea of how to delete all the media in the background, then delete the source without the request hanging because it is trying to delete the source instance.

meeb:
Nice, thanks!

PR #870, opened 2025-03-20
More stable `ffmpeg` releases

tcely:
This code picks the same build date, as long as:
1. The release still exists
2. The commit is the latest


meeb:
Yep seems fine to me, thanks for the effort of testing those jq args that looks painful.

PR #880, opened 2025-03-24
Change the task expiration to 1 day

tcely:
The node name for docker changes when creating a new container.

The PID can match after creating a new image if the start up sequence is the same.

~Using `docker restart` may still cause false positives for tasks still running.~

Used `boot_time` (from `/proc/kcore` mount) to determine if `locked_at` was before the current Docker start time.

### DIFF
--- a/patches/background_task/models.py
+++ b/patches/background_task/models.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
-from datetime import timedelta
+from datetime import datetime, timedelta, timezone as tz
 from hashlib import sha1
+from pathlib import Path
 import json
 import logging
 import os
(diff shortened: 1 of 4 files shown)


patches/background_task/models.py
```diff
@@ -38,6 +39,23 @@ def created_by(self, creator):
...
+
+    @property
+    def boot_time(self):
+        if self._boot_time > self.posix_epoch:
+            return self._boot_time
+        stats = None
+        boot_time = self.posix_epoch
+        kcore_path = Path('/proc/kcore')
+        if kcore_path.exists():
+            stats = kcore_path.stats()
```
meeb: This should be `kcore_path.stat()` shouldn't it?

tcely: Indeed.
```suggestion
            stats = kcore_path.stat()
```

tcely:
> Would this be more generic to use something like `psutil.boot_time()` ?

Unfortunately, that module gives the time the docker host was booted.

Here is an example from one of my containers:
```bash
$ grep '^btime' /proc/stat
btime 1742195491
```

```Python
>>> import os
>>> started = int(os.path.getmtime('/proc/kcore'))
>>> btime = 1742195491
>>> datetime.fromtimestamp(started) - datetime.fromtimestamp(btime)
datetime.timedelta(days=4, seconds=29539)
```

meeb:
Yep fair enough. Thanks!

meeb:
Ah, probably another small typo:

https://github.com/meeb/tubesync/actions/runs/14075926408/job/39418818396#step:6:168

tcely:
Yes. #883

patches/background_task/models.py
```diff
@@ -190,14 +208,23 @@ class Task(models.Model):
...
+    @property
+    def nodename(self):
+        return os.uname().nodename[:(64-10)]
+ 
     def locked_by_pid_running(self):
         """
         Check if the locked_by process is still running.
         """
-        if self.locked_by:
+        if self in objects.locked(timezone.now()) and self.locked_by:
```
tcely, line 219:
```suggestion
        if self in self.objects.locked(timezone.now()) and self.locked_by:
```

tcely: Applied this change in #883.

PR #886, opened 2025-03-26
Force new `yt-dlp` version

### LINKED ISSUE #882: yt-dlp nsig error fixed with 2025-03-26
I was experiencing an issue with all my audio files coming up as MISSING_POT due to an nsig error. I updated yt-dlp to 2025-03-26 and it fixed it. I don't have any logs from the previous error, just know that upgrading fixed it.

### LINKED ISSUE #887: "No matched source" when there is an exact match
I'm trying to download a number of videos and TubeSync tells me that there's no matched source. When checking on of the videos, TubeSync tells me there's an exact match.

Image

Image

Any idea what I can do to fix this?

Manually skipping & re-adding the media doesn't work.

I've also tried setting the video & audio codec to a different one, with the same result. Tried deleting & readding the source, also no go.

tcely:
Set the `YTDLP_DATE` build arg from the tag name of the latest release of `yt-dlp` using the GitHub API.

```
Run docker/build-push-action@v6
  with:
    platforms: linux/amd64,linux/arm64
    push: false
    tags: ghcr.io/tcely/tubesync:latest
    cache-from: type=registry,ref=ghcr.io/tcely/tubesync:latest
    cache-to: type=inline
    build-args: IMAGE_NAME=tubesync
  FFMPEG_DATE=2025-03-22-07-48
  FFMPEG_VERSION=N-118896-g9f0970ee35
  YTDLP_DATE=2025.03.26
  
    load: false
    no-cache: false
    pull: false
    github-token: ***
  env:
    IMAGE_NAME: tubesync
    FFMPEG_DATE: 2025-03-22-07-48
    FFMPEG_VERSION: N-118896-g9f0970ee35
```

Fixes #882
Fixes #887


meeb:
This file just needs a tag bump to use a new release every time now?

tcely:
It needs a change when `yt-dlp` is newer than the `ffmpeg` build and we want to upgrade.

PR #905, opened 2025-03-30
Add a post-processor to disallow partial playlist results

tcely:
The idea for this is simple.
Check that we download all of the entries the playlist reported.

Rarely, a playlist result is too short for whatever reason. When this happens, and you are using "delete removed media" then a lot of media items will have been deleted.

Later, when the full playlist results are retrieved, all of those media items are being downloaded again.

We want to avoid this situation by skipping the index when the playlist is incomplete.

```
[tubesync/DEBUG] [info] Writing updated playlist metadata as JSON to: /config/cache/yt-dlp/infojson/playlist/postprocessor_[PLID]_100_876_temp.info.json                                  
[tubesync/DEBUG] [Exec] Executing command: /usr/bin/env bash /app/full_playlist.sh 'PLID' '876'
```

Failed playlists look like this:
```
  File "/usr/local/lib/python3.11/dist-packages/yt_dlp/postprocessor/exec.py", line 30, in run
    raise PostProcessingError(f'Command returned error code {return_code}')
yt_dlp.utils.PostProcessingError: Command returned error code 1
```

### DIFF
--- a/tubesync/sync/youtube.py
+++ b/tubesync/sync/youtube.py
@@ -161,13 +161,13 @@ def get_media_info(url, days=None):
     postprocessors.extend((dict(
         key='Exec',
         when='playlist',
-        exec_cmd="/usr/bin/env sh /app/full_playlist.sh '%(id)s' '%(playlist_count)d'",
+        exec_cmd="/usr/bin/env bash /app/full_playlist.sh '%(id)s' '%(playlist_count)d'",
     ),))
-    infojson_directory_path = Path(opts.get('cachedir', '/dev/shm')) / 'infojson',
+    infojson_directory_path = Path(opts.get('cachedir', '/dev/shm')) / 'infojson'
     playlist_infojson = 'postprocessor_[%(id)s]_%(n_entries)d_%(playlist_count)d_temp'
     outtmpl = dict(
         default='',
-        pl_infojson=f'{infojson_directory_path!s}/playlist/{playlist_infojson}.info.json',
+        pl_infojson=f'{infojson_directory_path}/playlist/{playlist_infojson}.%(ext)s',
     )
     for k in OUTTMPL_TYPES.keys():
         outtmpl.setdefault(k, '')
--- a/tubesync/sync/youtube.py
+++ b/tubesync/sync/youtube.py
@@ -148,6 +148,7 @@ def get_media_info(url, days=None):
             f'yesterday-{days!s}days' if days else None
         )
     opts = get_yt_opts()
+    default_opts = yt_dlp.parse_options([]).options
     paths = opts.get('paths', dict())
     if 'temp' in paths:
         temp_dir_obj = TemporaryDirectory(prefix='.yt_dlp-', dir=paths['temp'])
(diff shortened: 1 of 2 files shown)


tubesync/sync/youtube.py
```diff
@@ -158,11 +158,11 @@ def get_media_info(url, days=None):
             'temp': str(temp_dir_path),
         })
     postprocessors = opts.get('postprocessors', default_opts.get('postprocessors', list()))
-    postprocessors.extend(dict(
+    postprocessors.extend((dict(
```
tcely, line 173: I swear Python tries to trip people up with things like this!

tubesync/sync/youtube.py
```diff
@@ -148,31 +148,65 @@ def get_media_info(url, days=None):
...
+        key='Exec',
+        when='playlist',
+        exec_cmd="/usr/bin/env bash /app/full_playlist.sh '%(id)s' '%(playlist_count)d'",
+    ),))
+    cache_directory_path = Path(user_set('cachedir', opts, '/dev/shm'))
+    playlist_infojson = 'postprocessor_[%(id)s]_%(n_entries)d_%(playlist_count)d_temp'
+    outtmpl = dict(
+        default='',
+        infojson='%(id)s.%(ext)s' if user_set('infojson', paths, '') else '',
```
tcely:
```suggestion
        infojson='%(id)s.%(ext)s' if paths.get('infojson') else '',
```

meeb:
Seems a reasonable feature... if pretty hacky, is this fully tested? Might be more sensible to store the temp data in a TempFile rather than on `/dev/shm` as well.

tcely:
> Seems a reasonable feature... if pretty hacky, is this fully tested?

It's hackish, and it's also doing what I needed on my test container.

The potential alternative is to patch `yt-dlp` yet again so that `n_entries` is set in `ie_result` instead of only `ie_copy`, which is used for generating file names, but not returned.

> Might be more sensible to store the temp data in a TempFile rather than on `/dev/shm` as well.

Why would a `TempFile` be preferable? I'm actually leaning towards always using `/dev/shm` for this.

meeb:
> It's hackish, and it's also doing what I needed on my test container.

OK, thanks.

> Why would a `TempFile` be preferable? I'm actually leaning towards always using `/dev/shm` for this.

Depends on the size of the data you're writing really, if it's small enough it's probably fine. Increasing the already pretty heavy RAM footprint of the container might not be ideal given this data is will likely also be stored in RAM by `yt-dlp` until the postprocessors finish if I understand it correctly from reading the flow. An extra few megabytes isn't worth worrying about but I've stumbled over massive JSON outputs previously with playlists.

tcely:
> > Why would a `TempFile` be preferable? I'm actually leaning towards always using `/dev/shm` for this.
> 
> Depends on the size of the data you're writing really, if it's small enough it's probably fine. Increasing the already pretty heavy RAM footprint of the container might not be ideal given this data is will likely also be stored in RAM by `yt-dlp` until the postprocessors finish if I understand it correctly from reading the flow. An extra few megabytes isn't worth worrying about but I've stumbled over massive JSON outputs previously with playlists.

That's a fair point about the RAM. We already use about 8 GiB in my most stressful configuration.

This particular file will exist for a short time, it's created by `yt-dlp` then removed by the post-processor script. However, that is enough time to increase the memory pressure.

I don't think it will actually increase our memory usage because it is gone before Python allocates memory to store the return value we use from `yt_dlp`, which should be the exact same data.

tcely:
Hah! I just found [this issue](https://github.com/yt-dlp/yt-dlp/issues/12759) which likely explains why I noticed this problem recently.

PR #920, opened 2025-04-03
Set priority in tasks.py

### LINKED ISSUE #535: Feature Request: Configurable Queue Priorities
Perhaps I'm using it wrong, but TubeSync seems to perform poorly when you add more than a handful of playlists/channels to it. 
I believe that, at least in part, this is caused by the Queue Priorities being not optimal. 

Currently it seems that actually downloading media is the last thing that will happen, so the task queue grows massively with the various tasks that need to be completed. 

What should happen, IMO is that the queue priorities should be inverted, or at least tweaked.  Downloading media that's been discovered should be the highest priority, followed by thumbnails and metadata, with the discovery tasks being lowest priority. 

Having the various task priorities read from env vars would do the trick, if you don't want to make this a global change for everyone. 

Would you accept a PR to make these priorities read from env vars?

### LINKED ISSUE #917: Configure task priorities in `tasks.py`
Tweaking task priorities is mostly where it needs to be.

The goal for this issue is to reduce the amount of coordination required in the `kwargs` when calling background tasks.

Fixes #535

tcely:
Fixes #535
Fixes #917

### DIFF
--- a/tubesync/sync/management/commands/delete-source.py
+++ b/tubesync/sync/management/commands/delete-source.py
@@ -6,7 +6,7 @@ from django.db.models import signals
 from common.logger import log
 from sync.models import Source, Media, MediaServer
 from sync.signals import media_post_delete
-from sync.tasks import rescan_media_server
+from sync.tasks import schedule_media_servers_update
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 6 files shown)


meeb:
Seems generally sensible to me, thanks!

Issue #47, opened 2021-02-17
Refactor sync models and rework metadata handling


meeb:
The `sync` app models are pretty unwieldy. They need a good tidy up and refactoring. Metadata should be split off and probably have some caching applied to avoid calling `json.loads(...)` excessively.

tcely:
I added some caching recently.

Splitting out the items from metadata, that we regularly use, then storing those all in a table would be a good next step.

tcely:
Playing around with an old `db.sqlite3` file this is what I used:

```sql
CREATE TABLE IF NOT EXISTS "sync_metadata" AS SELECT "metadata" -> '$.id' AS "key", "source_id", "published", datetime("metadata" -> '$.epoch', 'unixepoch') AS "updated", "metadata" AS "value" from "sync_media" ;

CREATE TABLE IF NOT EXISTS "sync_metadata_formats" ("metadata_id" REFERENCES "sync_metadata" ("rowid") ON DELETE CASCADE, "key" char(12) NOT NULL, "num" INTEGER NOT NULL, "format_id" varchar(20) NOT NULL, "value" json not null, UNIQUE("key", "num"), UNIQUE("key", "format_id"));

# key and format_id are duplicated data, but they are too useful for queries to eliminate

INSERT OR IGNORE INTO "sync_metadata_formats" ("metadata_id", "key", "num", "format_id", "value") SELECT "sync_metadata"."rowid", "sync_metadata"."key" AS "key", "each"."key" AS "num", "each"."value" ->> '$.format_id' AS "format_id", json_remove("each"."value", '$.url', '$.manifest_url', '$.fragments') AS "value" FROM "sync_metadata", json_each("sync_metadata"."value", '$.formats') AS "each" WHERE "each"."atom" IS NULL ORDER BY "sync_metadata"."updated" DESC;
```

tcely:
After merging #954, #968, and #992 I've started the data migration on my "stable" container.

![Image](https://github.com/user-attachments/assets/bbaf4a66-4ba5-43db-b1b3-2099ee4c98e8)

It's not going to be finished quickly, but there was no user involvement and even the health check is working.

I'm very happy see the memory hasn't increased and there hasn't been a slow down after processing a significant number of media items.

PR #968, opened 2025-04-16
Start using the new metadata models

tcely:
### Todo:

- [x] Finalize schema
- [x] Upload all migrations
- [x] Squash migrations
- [x] Adjust squashed migration file
- [x] Add data migration functions

For #47


meeb:
Not sure if you noticed but this seems to have broken the tests unfortunately.

tcely:
I did notice.

meeb:
This seems like a relatively major change, you happy for this to be merged? I've not tested it.

tcely:
No

tcely:
@meeb

After a review, this should be ok to merge.

The data migration needs attention, but I think that should just work. I tested the pieces it uses fairly often.

meeb:
Thanks! This seems like a big change that will be difficult to roll back if it breaks anything. It works for me, just re-confirming you're happy for this to merge?

tcely:
Yes, this is a large change. I do think it will be possible to remove it, if we needed to. However, I prefer not to need to.

It works for me, and I don't see anything that should cause it not to work for others.

Let's go ahead and merge.

Thanks for the review and confirmation.

PR #971, opened 2025-04-17
Add /etc/apt/apt.conf.d/docker-disable-pkgcache

tcely:
Read more at:
* https://github.com/debuerreotype/docker-debian-artifacts/issues/126
* https://github.com/debuerreotype/debuerreotype/pull/178


tcely:
I've added a few other build related changes, since this is going to invalidate the cached layers anyway when it is merged.

PR #976, opened 2025-04-19
Squash old migrations for `sync`

tcely:


tcely:
https://github.com/meeb/tubesync/issues/979#issuecomment-2819422193

When people try going from pre-30 versions to the latest, the migration files aren't there.

We may need to put 20-30, or 25-30, back into the directory.

PR #992, opened 2025-04-27
Use the actual model for data migration

tcely:
Well, that was a disappointing upgrade.

I remember reading about some limitations with the synthetic models, but I guess I didn't understand well enough.

We must use the real models to do the data migration.


tcely:
@meeb Please review and merge when you have a moment. Thanks.

PR #998, opened 2025-04-28
Add redownload thumb link

tcely:
I've wanted this in the UI for a bit of time now.

I'm welcoming changes to make it look better, but I at least want it to function correctly first.

For #997


meeb:
Nice, thanks!

PR #1001, opened 2025-04-29
Do not cache the blank thumbnail for so long

tcely:


meeb:
You can alternatively just stuff a `?last_updated=timestamp` query string on the end of thumbnail image `src` in the templates.

tcely:
I guess. This is cleaner for working in coordination with the caching done by `nginx`.

The query string approach just fills the cache with useless entries.

PR #1004, opened 2025-04-30
Include `missing_pot` formats in testing

tcely:


meeb:
Is this OK to merge without #1005 ?

tcely:
Yes. The client should be deprioritized and the formats should either work or not when they are tested.

Eventually, I want to provide an external way to generate these tokens via the plugin.

PR #1006, opened 2025-04-30
Replace `nginx` with `openresty`

tcely:


meeb:
What's this for?

tcely:
> What's this for?

It's a newer version, and includes features that I plan to use in the future.

Hopefully, the proof of origin stuff and dynamically controlled logging.

meeb:
Customised HTTP logging is fine. I don't really want any logic ported to lua that is required for tubesync to function.

tcely:
I was planning to use `lua` for the configuration. The goal is; if you aren't using `nginx` then nothing should change.

It really is a better approach than templates for generation of configuration before running `nginx`.

meeb:
Yep I'm very familiar with Openresty. Providing tubesync functions as a stand-alone Python/Django project and there's nothing functionally required in lua that's fine.

Issue #961, opened 2025-04-14
Update PostgreSQL database instructions


tcely:
The current [file](https://github.com/meeb/tubesync/blob/main/docs/other-database-backends.md) is outdated.
There are easier ways to accomplish the creation of a new database for use with `TubeSync` these days.

This is a great first issue to resolve!
It doesn't require knowing Python or the existing code base.

Items to address:

- [x] 1. The tag is too specific. [`15.2`](https://hub.docker.com/layers/library/postgres/15.2/images/sha256-8d0f925ceb8d8110a9d7c4f792fdabb50d797f5e3f25e21d059abf4fa177be2f) when just using [`15`](https://hub.docker.com/layers/library/postgres/15/images/sha256-f25135c3038e660f35edd6bc752769b67fa0c43dec1f691069f8089f6c2fd7a0) should be [`15.12`](https://hub.docker.com/layers/library/postgres/15.12/images/sha256-f25135c3038e660f35edd6bc752769b67fa0c43dec1f691069f8089f6c2fd7a0), as of this date.
- [x] 2. The [`latest`](https://hub.docker.com/layers/library/postgres/latest/images/sha256-48f04a5009fe444f00178907dd32f6df809246a959468f72248284752f31dadd) tag for [`PostgreSQL`](https://hub.docker.com/_/postgres) uses [`17`](https://hub.docker.com/layers/library/postgres/17/images/sha256-48f04a5009fe444f00178907dd32f6df809246a959468f72248284752f31dadd), so [`15`](https://hub.docker.com/layers/library/postgres/15/images/sha256-f25135c3038e660f35edd6bc752769b67fa0c43dec1f691069f8089f6c2fd7a0) is two versions behind, as of this date.
- [x] 3. Creating the database can be done with an environment variable ([`POSTGRES_DB`](https://github.com/docker-library/postgres/blob/cc254e85ed86e1f8c9052f9cbf0e3320324f0421/15/bookworm/docker-entrypoint.sh#L205)), instead of using the `init.sql` file.
- [x] 4. It's really a great idea to stop the gunicorn and worker services before importing the tasks. I typically don't let them run until after restarting the container.

See also:
* https://github.com/meeb/tubesync/issues/748#issuecomment-2673133029
* https://github.com/meeb/tubesync/issues/943#issuecomment-2801293005

PR #1016, opened 2025-05-05
Add more details about saving database entries

tcely:
For #961


tcely:
@meeb My almost 1 GiB SQLite database is backed up to 150 MiB file using `gzip`, but the Django support for `xz` further reduced that to less than 50 MiB.

It turned out that `jsonl` is extremely compression friendly. 👍

meeb:
Very nice!  Thanks!

PR #1030, opened 2025-05-09
Refactor models

### LINKED ISSUE #47: Refactor sync models and rework metadata handling

tcely:
The `Media` model is still a mess.
At least now it is a mess in its own file.

~I fully expect this to need a migration file.~
The generated migration file was added.

Fixes #47


tcely:
@meeb Let me know if you see anything we should change before merging this.

PR #1038, opened 2025-05-16
Migrate functions from `sync.utils`

tcely:


tubesync/common/utils.py
```diff
@@ -7,10 +7,21 @@
...
 
 
+def directory_and_stem(arg_path, /):
+    filepath = Path(arg_path)
+    stem = Path(filepath.stem)
+    while stem.suffixes and '' != stem.suffix:
+        stem = Path(stem.stem)
+    return (filepath.parent, str(stem),)
+
+
```
tcely, line 24: This was updated on `main` in the old location.

meeb:
Thanks!

PR #1042, opened 2025-05-17
Fixes for `ruff check` output

tcely:


tubesync/sync/views.py
```diff
@@ -296,23 +296,30 @@ class EditSourceMixin:
     def form_valid(self, form: Form):
         # Perform extra validation to make sure the media_format is valid
         obj = form.save(commit=False)
-        source_type = form.cleaned_data['media_format']
+        # temporarily use media_format from the form
+        saved_media_format = obj.media_format
+        obj.media_format = form.cleaned_data['media_format']
```
tcely, line 301: I only took a guess at what was intended here.

Setting the `source_type` variable with the `media_format` was very confusing.

https://github.com/meeb/tubesync/blob/08c1a82c30b403a48ed7a7433f2c9015738f6301/tubesync/sync/views.py#L293

https://github.com/meeb/tubesync/commit/08c1a82c30b403a48ed7a7433f2c9015738f6301#diff-7750fe54dfe02b74b11753719c6043d44f3a630604b97df81bc4336f9c5d50a0R293

@meeb Are there any memories about why it was added like this?

tubesync/sync/views.py
```diff
@@ -296,23 +296,30 @@ class EditSourceMixin:
...
                 ValidationError(self.errors['invalid_media_format'])
             )
 
         # Check for suspicious file path(s)
         try:
-            targetCheck = form.cleaned_data['directory']+"/.virt"
-            newdir = safe_join(settings.DOWNLOAD_ROOT,targetCheck)
+            targetCheck = form.cleaned_data['directory'] + '/.virt'
+            safe_join(settings.DOWNLOAD_ROOT, targetCheck)
         except SuspiciousFileOperation:
```
tcely, line 315: Using the exception to check for invalid values was cute, but not intuitive or obvious.

tubesync/tubesync/settings.py
```diff
@@ -222,5 +222,5 @@
     BACKGROUND_TASK_ASYNC_THREADS = MAX_BACKGROUND_TASK_ASYNC_THREADS
 
 
-from .dbutils import patch_ensure_connection
+from .dbutils import patch_ensure_connection # noqa
```
tcely, line 225: I considered moving this to the top of the file instead.

tubesync/sync/views.py
meeb: Given `source_type` isn't used in that method from what I can see I'd go with a copy/paste error that didn't break anything so went unnoticed.

PR #1075, opened 2025-05-27
Queues managed with `django-huey`

tcely:
Requirements:

- [django-huey](https://github.com/gaiacoop/django-huey)
- [huey](https://github.com/coleifer/huey)
- [gzip](https://docs.python.org/3.10/library/gzip.html#module-gzip)
- [sqlite3](https://docs.python.org/3.10/library/sqlite3.html#module-sqlite3)
- [Django](https://docs.djangoproject.com/en/5.1/ref/databases/#sqlite-notes) 5.1+
- [SQLite](https://www.sqlite.org/) 3.38.0+

```
#51 14.40 INFO:pipenv.patched.pip._internal.resolution.resolvelib.reporter:Reporter.adding
#51 14.40 _requirement(SpecifierRequirement('django-huey'), None)
#51 14.41 INFO:pipenv.patched.pip._internal.operations.prepare:Collecting django-huey 
#51 14.41 (from -r /tmp/pipenv-_zqkhb0y-requirements/pipenv-jzcdg_gx-constraints.txt (line
#51 14.41 16))
#51 14.42 INFO:pipenv.patched.pip._internal.network.download:Downloading 
#51 14.42 django_huey-1.2.1-py3-none-any.whl.metadata (5.1 kB)
```

The consumers should log any flushed locks at startup with:
`Found stale locks:`


tcely:
> Is this sensible to use over celery or the built in background tasks in Django 5.2?

Built-in background tasks? Please explain more. I saw nothing in the Django release notes.

As for if this is sensible, it is by far the closest match to what we need that I have seen so far.

tcely:
@meeb Please review this when you have a moment.

meeb:
Yep I'll take a look, thanks. This might be a while longer to review. I'll need to make sure I'm familiar with Huey (which I've never used previously) before merging.

tcely:
There shouldn't be anything using the `huey` queues yet.

I tried to break things up into logical steps to make things easier for all of us.

Having these changes in-place was the first step needed to experiment with tasks using the new system.

I've been playing around with `huey` and I am very impressed by how complete the coverage of common task cases is already.

Reading this file helped me to have a better understanding of how `huey` operates:

https://github.com/coleifer/huey/blob/master/huey/contrib/mini.py

Otherwise, the documentation was good enough in most places and where it wasn't the source code is easy to read for most of the things that I was curious about.

For my own future reference:
> Ah, [DEP 14](https://github.com/django/deps/blob/5746805dfb062a99555b5e5bbc5b1685212aba2c/accepted/0014-background-workers.rst) was accepted but the PR is still pending so it'll be in a future version of Django:

meeb:
Thanks, no issues after some time checking it out.

PR #1087, opened 2025-06-01
Enable `editor` in the container

tcely:
For less than 3 MiB, I can have a modern editor that does enough of what `vim-nox` does that I won't want to install another editor every time that I'm trying new things.

Also, for another 3 MiB we can get a replacement for `nano` that people newer to this command line editing experience can use more easily.

```sh
$ update-alternatives --get-selections
editor                         manual   /usr/bin/vis
nano                           auto     /usr/local/bin/babi
vi                             auto     /usr/bin/vis
vim                            auto     /usr/bin/vis
```


meeb:
In-container editor is nice, thanks!

PR #1132, opened 2025-06-18
Do not store `True` as a result

tcely:


meeb:
@tcely this has a conflict

tcely:
@meeb The conflict has been resolved.

PR #1153, opened 2025-06-24
Reorganize tasks

tcely:
We are down to 8 tasks using the old system.
They were all moved to the end of the file.

One of the workers was removed as no more tasks were using it.

Cleaning up the results:
```py
import django_huey
from pprint import pp
from sync.choices import Val, TaskQueue
pp( (lambda q, msg=None, remove_true=False: [ (n, k, v := q.get(peek=True, key=k), e := v.metadata.get('error') if hasattr(v, 'metadata') else v, 'Removed True' if remove_true and v is True and q.get(k) else 'Retained True', 'Removed' if msg and e and msg in e and q.get(k) else 'Retained') for n,k in enumerate(q.all_results(), start=1) ])(django_huey.get_queue(Val(TaskQueue.LIMIT)), ) )
```


meeb:
Nice work, thanks!

PR #1168, opened 2025-07-03
Add SQLite configuration tweaks

tcely:


tcely:
@spiralham Let us know what the timeout value that you used is. Thanks!

spiralham:
@tcely I just went with `60` which is probably overkill but seems to have cleaned up the lock errors on both btrfs and ext4.

PR #1177, opened 2025-07-05
Use an already verified and extracted docker image for `s6-overlay`

tcely:


meeb:
No objection to this, but unless you mind I may fork your fork just so that building tubesync doesn't depend on another account. Thoughts?

tcely:
I don't mind at all. Particularly if you add me as a collaborator to your fork.

meeb:
Done and added.

tcely:
I think you should be all set now.

https://github.com/meeb/docker-s6-overlay/pkgs/container/s6-overlay/versions?filters%5Bversion_type%5D=tagged

PR #1179, opened 2025-07-05
Use a `datetime.datetime` instead of a `tuple`

tcely:


tcely:
@meeb Please merge this as soon as you can.

PR #1180, opened 2025-07-05
Convert `delete_all_media_for_source` task

tcely:


tcely:
I just deployed the latest image and my memory is about 300 MiB less without that worker.

PR #1192, opened 2025-07-07
Tweaks to tasks for consistency

tcely:


tcely:
```
Getting action download info
Failed to resolve action download info. Error: The HTTP request timed out after 00:01:40.
Retrying in 14.861 seconds
```

That's interesting. I haven't noticed GitHub simply not responding before.

PR #1198, opened 2025-07-12
Revoke new duplicate tasks

### LINKED ISSUE #1194: Implement something like `remove_existing_tasks` for Huey tasks
The same task with the same `Media` instance should really only be scheduled or pending once for most cases.

Similarly, for `Source` instances we don't want multiple tasks in the pending list either.

Using the new `remove_duplicates` support:
```
TaskHistory.schedule(
    [TaskWrapper],
    str(media.pk),
    remove_duplicates=True,
    vn_fmt = _('Task with media for "{}"'),
    vn_args=(media.key,),
)
```

tcely:
Fixes #1194


tcely:
@meeb Please take a look when you have a moment.

PR #1228, opened 2025-07-17
Use media redownload to implement manual downloads

### LINKED ISSUE #384: [Feature request] Add manual download button
Tl;dr
Add a manual download button

With bigger or older yt channels, they might have several hundred videos. When adding them you don't always want all the videos (especially if you're a long time viewer) so when adding the source, you set the download threshold to 1 week. However, you look through their backlog (in the skipped media) and see some videos that you enjoyed in the past and want to fetch ad-hoc. There is currently no way to do this (other than click on the yt link) as the 1 week cut off date in tubesync is gospel.
Can you add a 'manual override' button that trumps other checks and just kicks off a download? That'd be nice, I think.

### LINKED ISSUE #1037: Selective Download
You can turn "Download Media" off for a source. I would have expected a "Download Now" button on media when this is set, but I can't find one. It would be useful to have it.

If you're interested, I might be able to implement this one if you can point me in the right direction.

tcely:
Fixes #384
Fixes #1037


meeb:
@tcely sorry, merge conflict.

tcely:
@meeb The conflicts have been resolved.

PR #1314, opened 2025-10-06
Avoid SIGKILL from `docker stop`

tcely:


meeb:
As hacks go I've seen worse :)

tcely:
So have I.

However, having a way to specify the expected time to gracefully shutdown the processes in the `Dockerfile` is a much cleaner solution, in my opinion.

Regardless, 10 seconds is a ridiculously short default timeout.

PR #1325, opened 2025-11-06
Auto detect source type

StefanCodes:
## What is this?

The current UI makes the user choose what kind of source they're adding during a validation step. Practically speaking, we don't need to user to provide this - the source type can be inferred from the URL.

This change simplifies the UI to accept a single URL, then figures out the supported source type (if any) by itself.

## How does it do that?

* Replaces the Create Source buttons with a single button that navigates to a more-generic Add Source page
* Updates the URL validator to apply each of the 3 supported validators on the URL to see if any pass
* If any validator passes, we use that source type and proceed. If none pass, it's not a valid YouTube playlist or channel URL and we reject it as before.
* This requires tightening down the channel validation rules, as we need to watch for reserved YouTube URLs like `/watch`, `/trending`, etc.

## Testing

* Updated the specs to pass with the new implementation

Down to a single button:
image

Updated validation page:
image

Validation fails on direct video link:
image

### DIFF
--- a/tubesync/sync/views.py
+++ b/tubesync/sync/views.py
@@ -189,7 +189,7 @@ class ValidateSourceView(FormView):
     template_name = 'sync/source-validate.html'
     form_class = ValidateSourceForm
     errors = {
-        'unsupported_format': _('URL does not match any supported format.'),
+        'invalid_url': _('That URL does not match any supported formats.'),
     }
     source_types = youtube_long_source_types
     validation_urls = youtube_validation_urls
@@ -222,7 +222,7 @@ class ValidateSourceView(FormView):
             except ValidationError:
                 continue
         # Source type wasn't detected - presumably it's not a valid URL
-        form.add_error('source_url', ValidationError(self.errors['unsupported_format']))
+        form.add_error('source_url', ValidationError(self.errors['invalid_url']))
         return super().form_invalid(form)
 
     def get_success_url(self):


meeb:
Thanks for the PR. This seems reasonable to me and a nice UX improvement. The reasons behind the add a source options being separated are long since legacy and have been removed so I don't have any objections to the change. I'll give it a test.

tubesync/sync/templates/sync/source-validate.html
```diff
@@ -1,22 +1,29 @@
...
-    <p>Example: <strong>{{ help_example }}</strong></p>
+    <h1>Add Source</h1>
+    <p>
+      Enter any YouTube channel or playlist URL below.
+    </p>
+    <p><strong>Supported formats:</strong></p>
+    <ul>
+      <li><strong>YouTube Channel:</strong> <code>https://www.youtube.com/@rossmanngroup</code></li>
+      <li><strong>YouTube Channel by ID:</strong> <code>https://www.youtube.com/channel/UCl2mFZoRqjw_ELax4Yisf6w</code></li>
+      <li><strong>YouTube Playlist:</strong> <code>https://www.youtube.com/playlist?list=PLkVbIsAWN2lsmovRO20_gtfUfgWi-XnnT</code></li>
```
tcely, line 16: If you are changing these URLs to static data, the current source of these URLs should be cleaned up, if possible.

tubesync/sync/templates/sync/source-validate.html
```diff
@@ -1,22 +1,29 @@
...
 
-{% block headtitle %}Validate a {{ help_item }}{% endblock %}
+{% block headtitle %}Add a YouTube Source{% endblock %}
 
 {% block content %}
 <div class="row no-margin-bottom">
   <div class="col s12">
-    <h1>Validate a {{ help_item }}</h1>
-    <p>{{ help_text|safe }}</p>
-    <p>Example: <strong>{{ help_example }}</strong></p>
```
tcely, line 10: Remove the setting of `help_item`, `help_text` and `help_example` from the context?

tubesync/sync/templates/sync/source-validate.html
```diff
@@ -1,22 +1,29 @@
...
+    <p><strong>Supported formats:</strong></p>
+    <ul>
+      <li><strong>YouTube Channel:</strong> <code>https://www.youtube.com/@rossmanngroup</code></li>
+      <li><strong>YouTube Channel by ID:</strong> <code>https://www.youtube.com/channel/UCl2mFZoRqjw_ELax4Yisf6w</code></li>
+      <li><strong>YouTube Playlist:</strong> <code>https://www.youtube.com/playlist?list=PLkVbIsAWN2lsmovRO20_gtfUfgWi-XnnT</code></li>
+    </ul>
   </div>
 </div>
 <div class="row">
-  <form method="post" action="{% url 'sync:validate-source' source_type=source_type %}" class="col s12 simpleform">
```
tcely, line 14: Remove `source_type` from the context?

tubesync/sync/views.py
```diff
@@ -189,9 +189,7 @@ class ValidateSourceView(FormView):
     template_name = 'sync/source-validate.html'
     form_class = ValidateSourceForm
     errors = {
-        'invalid_source': _('Invalid type for the source.'),
-        'invalid_url': _('Invalid URL, the URL must for a "{item}" must be in '
```
tcely, line 193: Is `invalid_url` no longer possible? If it still is, then why remove this at all?

tubesync/sync/views.py
```diff
@@ -213,15 +211,8 @@ def __init__(self, *args, **kwargs):
     def dispatch(self, request, *args, **kwargs):
         self.source_type_str = kwargs.get('source_type', '').strip().lower()
         self.source_type = self.source_types.get(self.source_type_str, None)
```
tcely:
```suggestion
```

tubesync/sync/views.py
```diff
@@ -213,15 +211,8 @@ def __init__(self, *args, **kwargs):
     def dispatch(self, request, *args, **kwargs):
         self.source_type_str = kwargs.get('source_type', '').strip().lower()
```
tcely:
```suggestion
```

.gitignore
```diff
@@ -140,3 +140,11 @@ Pipfile.lock
...
 # Ignore Jetbrains IDE files
 .idea/
+
+# Task queue DBs
+/tubesync/tasks/*.db
+/tubesync/tasks/*.db-shm
+/tubesync/tasks/*.db-wal
+
+# Compiled CSS
+tubesync/common/static/styles/*.css
```
tcely:
```suggestion
/tubesync/common/static/styles/*.css
```

tcely (changes requested):
Removing invalid tests doesn't seem wise.
We should still reject all the URLs we did before, correct?

tcely:
@meeb Please approve the workflow so that we can lint this pull request.

tubesync/sync/tests.py
tcely:
```suggestion
    YouTube_SourceType,
```

StefanCodes:
> Removing invalid tests seems unwise to me. We should still reject all the URLs we did before, correct?

I think this change is ok, but please check my math:

The [view](https://github.com/meeb/tubesync/pull/1325/files?w=1#diff-7750fe54dfe02b74b11753719c6043d44f3a630604b97df81bc4336f9c5d50a0R212) now iterates over the 3 source type validators.
The old UI took a source type as an arg and used that to decide how to validate URLs.  If I picked a source type of channel and then entered a playlist URL, the validator caught [that specific case](https://github.com/meeb/tubesync/pull/1325/files?w=1).
But that URL is valid for the [playlist validator](https://github.com/meeb/tubesync/blob/main/tubesync/sync/tests.py#L59), so in the new workflow, it would be valid.

The test cases removed are all for conditions where they don't match that particular source type validator, but they are valid for one of the other ones.

tubesync/sync/views.py
StefanCodes: That's a good point. I replaced the old message with a more generic one (since we aren't exposing source type to the user anymore), but I didn't need to add a new key.
Updated in https://github.com/meeb/tubesync/pull/1325/commits/b9c404d5b60a0291db4a24f31695721dc15136fc

tcely:
> The test cases removed are all for conditions where they don't match that particular source type validator, but they are valid for one of the other ones.

That makes sense to me. Thanks.

tcely (approved):

meeb:
Tested and works well. Thanks for the contribution!

PR #1362, opened 2026-01-12
Run from a parallel directory

tcely:


meeb:
Just curious but how big could `/dev/shm/yt-cipher` get?

tcely:
It's less than 40K for that entire mount point.

```
Filesystem      Size  Used Avail Use% Mounted on
tmpfs            64M     0   64M   0% /dev
shm              64M   40K   64M   1% /dev/shm
```

`cp` is creating directories containing symbolic links to the files for the most part.

```
$ (cd /dev/shm/yt-cipher/ && find . '!' -type d '!' -type l -ls )
70     16 -rw-r--r--   1 app      app         13644 Jan 12 13:26 ./deno.lock
68      4 -rw-r--r--   1 app      app            94 Jan 12 13:26 ./deno.json
```

The actual data the server writes, if you use it at all, is expected to live in the cache.

```
Cleaning up player cache directory: /config/cache/yt-cipher/player_cache
Player cache directory ensured at: /config/cache/yt-cipher/player_cache
```

I've left a `yt-cipher` container active, and in use, for a few days now. It looks like every player file it has cached is ~2.6MiB.

PR #1363, opened 2026-01-15
Set `raise_incomplete_data` to cause task retries

tcely:
https://github.com/yt-dlp/yt-dlp/blob/27afb31edc492cb079f9bce9773498d08e568ff3/yt_dlp/extractor/youtube/_base.py#L1257


tcely:
@meeb I'm not sure I understand this matrix thing. Did you get a set of emojis on your end?

meeb:
Yeah, see email. You need to verify they are the same OOB.

PR #1372, opened 2026-01-30
Implement `revoke-task` feature

### LINKED ISSUE #1292: Aborting tasks
I clicked 'Reset Tasks' without being fully aware what it does, hoping that it would restart a seemingly suck download task.
Now, my whole library has been scheduled to be reindexed, renamed etc. which I didn't intend to do.
Is there any way to cancel these tasks or do I just have to let them run through?

tcely:
Closes #1292


meeb:
Nice, thanks.

Issue #1366, opened 2026-01-29
Sponsorblock segments not being removed


IAM-marco:
As mentioned in https://github.com/meeb/tubesync/issues/109#issuecomment-3417856539 , I'm experiencing the same issue, where the sponsorblock segments are not being removed, just marked through chapters.

Here are the settings of my video source:

<img width="1231" height="530" alt="Image" src="https://github.com/user-attachments/assets/538e11b0-961e-4f3c-9e4c-5c15708d3bdf" />

Here's the logs:

```
tubesync  | 2026-01-29 20:07:57,883 [tubesync/INFO] Downloading media: myzE710ehCA (UUID: 4b7d8983-dc9f-46d5-ba23-110ffcded0cd) to: "/downloads/video/Watch Later/Baldly Rudy S2025E0829 - The Ultimate Guide for Unlocking More Frames on the Steam Deck.mkv"
tubesync  | 2026-01-29 20:07:57,985 [tubesync/INFO] [youtube-dl] using cookies.txt from: /config/cookies.txt
tubesync  | 127.0.0.1 - - [29/Jan/2026:20:08:21 +0100] "GET /ping HTTP/1.1" 502 556 "-" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/139.0.0.0 Safari/537.36"
tubesync  | 2026/01/29 20:08:21 [error] 185#185: *85 connect() failed (111: Connection refused) while connecting to upstream, client: 127.0.0.1, server: _, request: "GET /ping HTTP/1.1", upstream: "http://127.0.0.1:4406/ping", host: "127.0.0.1:4416"
tubesync  | WARNING: [youtube] [pot:bgutil:http] HTTP Error reaching GET /ping (caused by <HTTPError 502: Bad Gateway>)
tubesync  | 2026-01-29 20:08:24,803 [tubesync/INFO] [SponsorBlock] started for: myzE710ehCA: The Ultimate Guide for Unlocking More Frames on the Steam Deck
tubesync  | 2026-01-29 20:08:24,819 [tubesync/INFO] [SponsorBlock] started for: myzE710ehCA: The Ultimate Guide for Unlocking More Frames on the Steam Deck
tubesync  | 2026-01-29 20:08:24,957 [tubesync/INFO] [SponsorBlock] finished for: myzE710ehCA: The Ultimate Guide for Unlocking More Frames on the Steam Deck
[...]

PR #1373, opened 2026-01-31
Expand selected_choices

tcely:
When every choice was selected we condensed that down to the `all_choice`.

We need to expand from `all_choice` to include every possible choice for `yt-dlp` to match removals properly.

For #1366


meeb:
Makes sense, thanks.

PR #1376, opened 2026-02-01
Improve `/app/restart_services.sh`

tcely:
- ~`s6-rc-db` has a useful list of bundles~
- `sc-rc` list accepts bundles/services the same way
- ~`s6-rc-db` can expand multiple bundles~
- `s6-rc-db` reports the type of a service
- Add `is_a_longrun` & `only_longruns` functions
- By default, restart all `longrun` services that are already `up`
- Do not try to restart essential services
- Remove find/grep work-arounds
- Remove the check for a bundle
- Remove the list of longruns and seevices

### DIFF
--- a/tubesync/restart_services.sh
+++ b/tubesync/restart_services.sh
@@ -26,7 +26,7 @@ only_longruns() {
 
 if [ 0 -eq $# ]
 then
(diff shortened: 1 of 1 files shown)


meeb:
No comments from me, merge away when you're finished tweaking.

PR #1382, opened 2026-02-05
Clean up outdated settings

tcely:
Removed:

- `MAX_ATTEMPTS`
- `BACKGROUND_TASK_PRIORITY_ORDERING`
- `MAX_BACKGROUND_TASK_ASYNC_THREADS`
- `BACKGROUND_TASK_ASYNC_THREADS`
- `BACKGROUND_TASK_RUN_ASYNC`
- `DEFAULT_THREADS`

### DIFF
--- a/tubesync/tubesync/local_settings.py.container
+++ b/tubesync/tubesync/local_settings.py.container
@@ -15,10 +15,6 @@ if DJANGO_URL_PREFIX and STATIC_URL:
     STATIC_URL = urljoin(DJANGO_URL_PREFIX, STATIC_URL[1:])
 
 
-# This is not ever meant to be a public web interface so this isn't too critical
-SECRET_KEY = getenv('DJANGO_SECRET_KEY', 'tubesync-django-secret')
-
-
 ALLOWED_HOSTS_STR = getenv('TUBESYNC_HOSTS', '*')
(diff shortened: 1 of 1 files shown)


tubesync/tubesync/local_settings.py.container
```diff
@@ -15,10 +15,6 @@ if DJANGO_URL_PREFIX and STATIC_URL:
     STATIC_URL = urljoin(DJANGO_URL_PREFIX, STATIC_URL[1:])
 
 
-# This is not ever meant to be a public web interface so this isn't too critical
```
meeb, line 18: This probably needs to left in - pretty sure Django doesn't start without it being set to something.

tcely: The old state has settings with '', which was overridden by local_settings with the environment variable, defaulting to 'tubesync-django-secret'.

The new state is to read the environment variable directly in settings, then not override that value.

meeb: I personally don't like the default `settings.py` loading env vars at all, I generally like `settings.py` to always be static and then have `local_settings.py` override anything that might be dynamic. Makes "where the dynamic config lives" clearer. Personal preference though, and really not that important for this particular example so fine with this as-is to avoid a debate on a very minor issue.

PR #1386, opened 2026-02-11
Improve `get_histories_from_huey_ids`

tcely:
Revise syntax and types

Use more universally accepted syntax and tighten the types.


meeb:
Fine with me, merge away when ready.

PR #1412, opened 2026-03-02
Reconfigure debconf for Docker

tcely:
Clean up the major dive finding by treating the image file as a read-only component of the database.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,19 @@ ARG S6_CHECKSUM_ALGORITHM="sha256"
 ARG QJS_CHECKSUM_ALGORITHM="sha256"
 
 
+FROM debian:${DEBIAN_VERSION} AS tubesync-prepare-etc
+
+COPY patches/ /var/tmp/patches/
+RUN --mount=type=tmpfs,target=/cache \
+    set -eux && cd /var/tmp/patches/ && \
+    ./fettle.pl --dry-run ./docker/tubesync-base/debconf.diff && \
+    ./fettle.pl ./docker/tubesync-base/debconf.diff && \
+    ./fettle.pl --clean ./docker/tubesync-base/debconf.diff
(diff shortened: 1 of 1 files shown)


meeb:
Can you clarify what this actually does? Why is it particularly matter if the Debconf templates are write-able at container build time?

tcely:
> Can you clarify what this actually does? 

Sure.

> Why is it particularly matter if the Debconf templates are write-able at container build time?

```
Analyzing image...
  efficiency: 99.2192 %
  wastedBytes: 6650427 bytes (6.7 MB)
  userWastedPercent: 1.0326 %
Inefficient Files:
Count  Wasted Space  File Path
    5        4.1 MB  /var/cache/debconf/templates.dat
    2        742 kB  /var/cache/debconf/templates.dat-old
```

If you look at the current dive analysis above, you can see that the flat file database is present in multiple layers and occupies the top spot for wasted space.

By treating the file in the image as read-only we can eliminate this churn of a single file every time debconf updates a template.

Instead of overwriting the file multiple times, individual files per package are created for the new content.

This keeps the layers clean and removes the wasted space from old versions of the flat file database.

meeb:
Ah, thanks. How much space does this save? If it's obvious in the analysis I'll check it myself in a bit.

tcely:
> Ah, thanks. How much space does this save? If it's obvious in the analysis I'll check it myself in a bit.

It's not a lot. Fixing this was simply on my to-do list.

I have added the dive analysis lines to my earlier comment.

tcely:
> The big Perl script is a bit overkill but sure.

It is. The nice thing about it, is that I can continue to use that on any Debian images, without requiring multiple steps to ensure that `patch` is available.

For this case, it was important to not use `debconf`, even indirectly, before the configuration was changed.

PR #1447, opened 2026-04-11
feat: generate health check URL

### LINKED ISSUE #1446: Change health check to generate its URL
We have environment variables for where `gunicorn` is listening, so we should use those to generate the proper URL instead of using a hard-coded argument in the `Dockerfile`.

I'd suggest instead of an error, we should format a string as the URL instead. This preserves the ability to override the URL with an argument but uses the generated one by default.

https://github.com/meeb/tubesync/blob/33479e8e36a23320e2107e42e051b7c4b5264c99/tubesync/healthcheck.py#L38-L42

https://github.com/meeb/tubesync/blob/33479e8e36a23320e2107e42e051b7c4b5264c99/tubesync/tubesync/gunicorn.py#L19-L24

https://github.com/meeb/tubesync/blob/33479e8e36a23320e2107e42e051b7c4b5264c99/Dockerfile#L776-L778

tcely:
Fixes #1446

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -774,7 +774,7 @@ RUN set -x && \
   printf -- "qjs_version = '%s'\n" "${qjs_version}" >> /app/common/third_party_versions.py
 
 # Create a healthcheck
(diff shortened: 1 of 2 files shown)


meeb (approved):
Nice, thanks!

You and your Yoda logic :)

PR #1448, opened 2026-04-11
feat: check the latest `yt-dlp` version in healthcheck

tcely:
Stop the `yt-dlp` queue from executing tasks when we notice that there is a new latest release at: https://github.com/yt-dlp/yt-dlp/releases/latest

### DIFF
--- a/tubesync/healthcheck.py
+++ b/tubesync/healthcheck.py
@@ -60,7 +60,7 @@ def bg_v(s, lock):
             now = time.time()
             atomic_write(s, v)
             atomic_write(s + '.t', now)
-            atomic_write(s + '.i', random.randint(1200, 21600))
+            atomic_write(s + '.i', random.randint(1200, 43200))
     except:
         pass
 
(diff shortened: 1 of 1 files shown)


meeb (approved):
Nice, thanks!

If I was nitpicking maybe put a fuzz of +/- a random hour on the local yt-dlp version cache file expiry? Not that I think it would realistically be a problem, but managed platforms like Unraid might restart all containers at a specific time or something weird. Up to you, very much on the finer end of the nitpicking scale.

tcely:
The time between checks is already randomized between 20 minutes and 6 hours. When the update happens, another random interval to the next check is chosen.

Does that address the concern sufficiently?

meeb:
> The time between checks is already randomized between 20 minutes and 6 hours. When the update happens, another random interval to the next check is chosen.
> 
> Does that address the concern sufficiently?

Yep! Honestly just missed the `randomint` when skimming it. A mean check time of just over 3 hours might be a little excessive, an average of once a day should be fine? As mentioned though nitpicking, whatever you think is reasonable here is fine.

PR #1449, opened 2026-04-12
chore(gunicorn): review and update configuration

tcely:
https://gunicorn.org/reference/settings/


meeb (approved):
Yep looks good, thanks.

PR #1452, opened 2026-04-13
fix(matching): prefer original audio in combined format

### LINKED ISSUE #1451: How to download the original audio instead of the default
Hi @tcely 

I'm having an issue with the audio in some downloads.

For example, with the video in this link: https://www.youtube.com/watch?v=E-3ezKJeNdo, it's downloading the audio in english, generated by AI instead of the original audio of the youtuber, spanish. The english audio seems to be label as default, but I want the original audio version.

Is there anyway to force tubesync to download the original instead of the default audio?

I'm using Tubesync version 0.17.1 with yt-dlp version 2026.03.17

Here's some screenshots for this example:

Image

Image

Image

Image

tcely:
Fixes #1451

### DIFF
--- a/tubesync/sync/matching.py
+++ b/tubesync/sync/matching.py
@@ -73,12 +73,19 @@ def get_best_combined_format(media):
         matches.add(fmt['id'])
         by_fmt_id[fmt['id']] = fmt
         by_language[fmt['language_code']] = fmt['id']
-        if 'format_note' in fmt and '(default)' in fmt['format_note']:
-            by_fmt_id['default'] = fmt
+        if 'format_note' in fmt:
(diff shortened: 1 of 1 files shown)


tcely:
@meeb 

Do you think it is worth adding a setting to consistently choose either original or default formats?

meeb (approved):

meeb:
> @meeb
> 
> Do you think it is worth adding a setting to consistently choose either original or default formats?

Yes that sounds like a nice per-source toggle to have available. It's probably more of a documentation challenge rather than the feature itself to accurately communicate what it does to people.

This is fine to merge though from what I can see. Thanks!

PR #1465, opened 2026-04-19
fix: lowercase `Source.name` before sorting

### LINKED ISSUE #1464: Source-list sorting by ASCII-code?
I've noticed that my source-list is sorted in a weird way. It looks like this:
```
Aa
Bb
...
Yy
Zz
aa
bb
...
yy
zz
```
I very much expect `mi...` to come before `My...`, not after `Zz...`.

tcely:
`PostgreSQL` was already being reasonable.
This should provide the same sorting for `SQLite` as well.

Fixes #1464


meeb (approved):
Yep nice, thanks.

PR #1472, opened 2026-04-27
feat(logs): switch to `hat-syslog`

### LINKED ISSUE #70: Include log in /config folder

### LINKED ISSUE #74: User-configurable log levels
Some of the log events are pretty spammy. While they are logged as `debug` events that's the default log level. Users should be able to sent a `LOG_LEVEL` style env var to limit log spam if they want.

tcely:
Fixes #70 #74
Replaces #713 

I found [hat-syslog](https://github.com/hat-open/hat-syslog) which looks like an excellent way to handle logs from `nginx` and potentially anything else that uses a `syslog` server.

https://hat-syslog.hat-open.com/server.html

- `--log-level 'INFO'`
- `--db-path "${XDG_STATE_HOME}/hat/syslog.db"`
- `--db-enable-archive`

https://github.com/tcely/hat-syslog/raw/refs/heads/patch-3/src_py/hat/syslog/older_formats.py

--------

https://gunicorn.org/reference/settings/#logconfig_dict

https://github.com/benoitc/gunicorn/blob/9bc5891b4b06f25a8ce0e707053dcb2fb9bf638c/gunicorn/glogging.py#L47-L86

https://github.com/benoitc/gunicorn/blob/master/gunicorn/glogging.py

--------

```
ghcr.io/meeb/tubesync:test@sha256:1914257ede1262b1a0f43fff8e11f17b00cce4a988724d0784c978a85bafb407
```

Screenshot


meeb (commented):
Thanks, this seems like quite a lot of work. Just so I can fully understand what you want to switch to here what's the benefit of embedding our own syslog server over just writing logs to stdout of the container? Also what does busybox-syslogd do here? If this is merged are logs still sent to stdout but there's also an HTTP interface to logs embedded in the container with all internal services hubbing around hat-syslog? Is this just to make logs conform to a defined format from all the embedded processes?

The review for this is being complicated by GitHub just failing to properly display or render the changed files. I've tried a few times but it only shows 2-3 changes per page load so I've resorted to git diffs.

tcely:
> Thanks, this seems like quite a lot of work.

It was. Hopefully, reviewing it will be less troublesome.

> Just so I can fully understand what you want to switch to here what's the benefit of embedding our own syslog server over just writing logs to stdout of the container?

The syslog server was the original idea based on your comments about rotation and my desire to easily filter the logs.

This particular implementation writing to a database and providing a web interface for viewing logs seemed to provide all the capabilities we both wanted.

We can log much more and configure the console to be as clean as we like or leave the default logging and ignore the records rotating through the table entirely.

> Also what does busybox-syslogd do here?

We needed a `/dev/log` which wasn't provided by the `pthyon` syslog endpoint.

That's provided by `busybox-syslogd` because it's small and still handles both rotated logs and forwarding to a remote syslog server. Both web servers are now logging through this path. Other things that can speak to `/dev/log` could also use this standard syslog capability.

https://github.com/meeb/tubesync/blob/6bd08ee8a3b8541b7b321b38afb520c1b6576b82/config/root/etc/s6-overlay/s6-rc.d/busybox-syslogd/run#L14-L25

> If this is merged are logs still sent to stdout but there's also an HTTP interface to logs embedded in the container with all internal services hubbing around hat-syslog? Is this just to make logs conform to a defined format from all the embedded processes?

Logs (from Python) are currently configured to go to multiple handlers. Web logs are sent to `/dev/log` only, but they are written to a rotated file in the container and forwarded to the syslog endpoint and stored in the database the web interface uses also.

A large part of the enhancement is to have all logs in a place we can filter / search that survives when a container is removed.

> The review for this is being complicated by GitHub just failing to properly display or render the changed files. I've tried a few times but it only shows 2-3 changes per page load so I've resorted to git diffs.

The web interface is quite bothersome with this change set. I recommend pulling the test image and inspecting the actual results as well as using the git CLI tools.

meeb:
Yep that's pretty much what I had assumed was going on, thanks for the confirmation. Provided we still forward all logs to stdout by default (I'm assuming hat-syslog can relay to stdout as an output or similar?) so general container logging systems don't break for everyone I've no objections.

Thanks!

tcely:
> Provided we still forward all logs to stdout by default (I'm assuming hat-syslog can relay to stdout as an output or similar?) so general container logging systems don't break for everyone I've no objections.

The logging is handled by the standard Python logging system, so the stream handler is still in there receiving a copy. We added another handler that creates records that are ingested by the syslog endpoint for storage.

https://github.com/meeb/tubesync/blob/6bd08ee8a3b8541b7b321b38afb520c1b6576b82/tubesync/tubesync/settings.py#L128-L154

I've been fairly happy with it so far, but of course we'll see what other people think after it's merged. 🙂

meeb:
If this eats HTTP logs it's going to annoy people that have HTTP log container analysis systems in place to monitor the health of their applications. Dual-streaming Django logs is nice but there's a bunch of other processes that run that aren't just Django.

Any way to just default to having hat-syslog relay logs to stdout and have an env var for people to turn it off if they don't want that? Otherwise this is modifying what is logged to stdout for everyone which was basically my only issue with the original discussion on not logging HTTP requests.

For example, something weird happening with gunicorn causes an HTTP/500 to be raised which is a clear and generally understood way of detecting issues.

As a feature this is nice, I'm just conservative about changing how things work regarding logging given the baseline standard for containers is just spam things to stdout and use another platform to store and parse them.

Providing this doesn't eat logs and the default is still to log to stdout (with any options you might want to disable that if that's your preference) I have no problems with merging it. Basically I'm just referring to changing the logging targets for gunicorn and Openresty if these are no longer passed on to stdout, stream stuff with Django is great.

tcely:
> If this eats HTTP logs it's going to annoy people that have HTTP log container analysis systems in place to monitor the health of their applications. Dual-streaming Django logs is nice but there's a bunch of other processes that run that aren't just Django.
> 

The good news is that it doesn't eat the `gunicorn` error logs. We simply added an additional handler to its logger.

> Any way to just default to having hat-syslog relay logs to stdout and have an env var for people to turn it off if they don't want that? Otherwise this is modifying what is logged to stdout for everyone which was basically my only issue with the original discussion on not logging HTTP requests.
>

This is thinking about the flow in the wrong direction. `hat-syslog` is an endpoint. To get copies we either configure the source differently or have an intermediary like `busybox-syslogd` copy the message to multiple locations for us.


> For example, something weird happening with gunicorn causes an HTTP/500 to be raised which is a clear and generally understood way of detecting issues.
>

This is how the error log operates now even with the additional syslog handler.

~We can easily add an additional error log line for `nginx` also.~ I forgot doing that already.

> As a feature this is nice, I'm just conservative about changing how things work regarding logging given the baseline standard for containers is just spam things to stdout and use another platform to store and parse them.
> 
> Providing this doesn't eat logs and the default is still to log to stdout (with any options you might want to disable that if that's your preference) I have no problems with merging it. Basically I'm just referring to changing the logging targets for gunicorn and Openresty if these are no longer passed on to stdout, stream stuff with Django is great.

Great! ~I'll adjust the `nginx` configuration shortly.~

https://github.com/meeb/tubesync/blob/4cd02ed690bd0664586cafe0e128802e0927683b/config/root/etc/nginx/nginx.conf#L66-L69

--------

Messages stored inside the container are already useful. 😉

```
May 23 12:32:30
  nginx: 2026/05/23 12:32:30
    [notice] 249#249: *607 a client request body is buffered to a temporary file /usr/local/openresty/nginx/client_body_temp/0000000212, client: 127.0.0.1, server: _, request: "POST /get_pot HTTP/1.1", host: "127.0.0.1:4416"
```

meeb (approved):
Thanks for the chat. I realise hat-syslog is the endpoint, however many syslog-based services support forwarding themselves anyway.

All good though, your adjustments to add back in some additional logs address my comments above.

PR #1486, opened 2026-05-27
feat: implement `PeekableSimpleQueue`

tcely:


meeb (commented):
No issue with the code, what's this for long term?

tcely:
I was looking at using it inside the logging thread but there just aren't enough upsides.

meeb:
Are there logging bottlenecks or other concerns that need something this extensive? If you want to just add it because it's cool that's fine as well.

tcely:
I liked the idea of the consistency of using the same interface for both the incoming and internal queues.

Instead of switching between `queue.Queue` methods and `collections.deque` methods, we could have stuck to the same methods because `queue.SimpleQueue` was designed to be compatible with `queue.Queue` with less overhead.

The only operation that we were doing that wasn't supported was looking at the first item and attempting to send it before removing it from the internal queue.

That alone simply wasn't enough reason to switch to importing this extended class. There's certainly no bottleneck that I am aware of.

meeb:
Sure, fine with me.

PR #1510, opened 2026-06-14
chore: convert full_playlist.sh

tcely:
* chore: convert full_playlist.sh  
Use a Django management command so that we can access settings and use the ORM.
* feat: check for configured sources during playlist validation  
Allow indexing partial playlist results when we can positively identify that it does not delete any missing videos.

TODO:

- [x] Check `Source` settings


meeb (approved):
Much nicer, thanks!

PR #1565, opened 2026-08-24
feat: use indexing response to save channel images

### LINKED ISSUE #1550: Rework source images to depend on indexing instead
Currently there is duplication.

~~For new sources we are essentially running two indexing tasks.~~ After #1561 was merged, we are downloading far less than the normal indexing task.

https://github.com/meeb/tubesync/blob/7ee8c9667ba7eeb7537c3044d578e94d54e5e583/tubesync/sync/models/source.py#L484-L488

https://github.com/meeb/tubesync/blob/7ee8c9667ba7eeb7537c3044d578e94d54e5e583/tubesync/sync/youtube.py#L83-L148

We should instead index the source, moving the information needed for downloading images into the database immediately, then rework the download of source images to wait for the data to become available from the database.

https://github.com/meeb/tubesync/blob/7ee8c9667ba7eeb7537c3044d578e94d54e5e583/tubesync/sync/tasks.py#L506-L689

https://github.com/meeb/tubesync/blob/7ee8c9667ba7eeb7537c3044d578e94d54e5e583/tubesync/sync/models/source.py#L559-L603

https://github.com/meeb/tubesync/blob/7ee8c9667ba7eeb7537c3044d578e94d54e5e583/tubesync/sync/youtube.py#L171-L279

ryanduguid:
Closes #1550.

## What

Reworks source-image downloads to depend on data already returned by source indexing:

- `Source.get_index()` pops `entries` and stores the remaining source-level response in the existing `Metadata` model, keyed by the videos/streams index URL.
- Valid stored metadata values, including thumbnail URLs, are retained when a later response contains missing or empty defaults.
- `download_source_images` looks up that indexed metadata and raises Huey's `CancelExecution` while it is unavailable, using the task's configured retries instead of a second yt-dlp extraction.
- No `Source` URL columns or schema migration are added.

## Verification

Regression coverage includes:

- source-level index responses are stored without media `entries`
- valid stored values survive incomplete later responses
- source images come from indexed metadata, not a second extraction
- unavailable or incomplete indexed data causes the image task to retry


tcely:
I haven't looked at your code yet, but I wanted to thank you for attempting this and point out a way to avoid the migration.

There is already a `Metadata` class that you can use. I was thinking of creating an entry for the `Source` instance with `NULL` for the `Media` that has its `Metadata.value` set to contain the response parts that we want to preserve from indexing.

`md = Metadata(source=self, key=self.key, media=None, value=(response|dict(entries=[])))` or something like this.

tubesync/sync/tasks.py
```diff
@@ -734,7 +734,12 @@ def download_source_images(source_id):
         log.error(f'Task download_source_images(pk={source_id}) called but no '
                   f'source exists with ID: {source_id}')
         raise CancelExecution(_('no such source'), retry=False) from e
-    avatar, banner, thumbnail = source.get_image_url
+    avatar = source.avatar_image_url
+    banner = source.banner_image_url
+    thumbnail = source.thumbnail_image_url
+    if not any((avatar, banner, thumbnail)):
+        # No URLs stored yet; fall back to a direct extraction.
+        avatar, banner, thumbnail = source.get_image_url
```
tcely: We should let the task be retried later instead of falling back to another use of `yt_dlp`.

`raise CancelExecution(_('data not yet available'))`

[CancelExecution](https://huey.readthedocs.io/en/latest/guide.html#canceling-from-within-a-task)

tcely (changes requested):
What you put together all looks reasonable.

I do want to point out that when we have good values in the database we don't necessarily want to update them with default values for responses that may be broken.

Also, see my earlier comments about using the tasks system instead of falling back to the old way and consider not using a migration.

Thank you for contributing!

tubesync/sync/models/source.py
```diff
@@ -574,6 +635,9 @@ def get_index(self, url_type, /):
         else:
             if not isinstance(response, dict):
                 return entries
+            # Keep the channel image URLs from this response so downloading
+            # source images does not need a separate extraction later.
+            self.save_indexed_image_urls(response)
             entries = response.get('entries', list())
```
tcely:
```suggestion
            entries = response.pop('entries', list())
            from .metadata import Metadata
            md, _ = Metadata.objects.defer('value').filter(
                media__isnull=True,
            ).get_or_create(
                source=self,
                site=response['extractor_key'],
                key=response.get('id', self.key),
            )
            field_defaults = {
                f.attname: f.get_default()
                for f in md._meta.fields
                if f.has_default()
            }
            if 'retrieved' in field_defaults:
                md.retrieved = field_defaults['retrieved']
            md.value = response
            md.save()
```

By first extracting `entries`, we are then left with the rest of `response`. We can now store that into the `Metadata` table for any number of later tasks to refer to.

tubesync/sync/models/source.py
```diff
@@ -574,7 +590,39 @@ def get_index(self, url_type, /):
         else:
             if not isinstance(response, dict):
                 return entries
-            entries = response.get('entries', list())
+            response = response.copy()
```
tcely:
```suggestion
```

This seems pointless.

tubesync/sync/models/source.py
```diff
@@ -19,12 +19,19 @@
...
+    get_image_urls as get_youtube_image_urls,
+    merge_image_thumbnails,
 )
 from ._migrations import media_file_storage
 from ._private import _srctype_dict
 
 
+def _has_index_metadata_value(value):
+    return value is not None and (
+        not isinstance(value, (str, list, dict)) or bool(value)
```
tcely:
```suggestion
        not isinstance(value, (str, list, dict, type(None))) or bool(value)
```

You don't need to check for `None` before using `isinstance`.

You can actually simplify this quite a bit:

```py
def _has_index_metadata_value(value):
    return bool(value) or value in (0, False,)
```

tubesync/sync/tasks.py
ryanduguid: Done. `download_source_images` now raises `CancelExecution(_('data not yet available'))` when indexed source metadata is missing or has no image URLs, so Huey retries instead of running another yt-dlp extraction.

tubesync/sync/models/source.py
ryanduguid: Done, with two small adjustments so later incomplete index responses do not replace good stored data:

- The source-level `Metadata` row is keyed by the index URL (`videos` / `streams`) and looked up with `media__isnull=True`.
- Existing non-empty values, including thumbnail URLs, are merged into the new response before save.

`download_source_images` then reads those rows instead of adding Source URL columns.

tubesync/sync/models/source.py
```diff
@@ -574,7 +590,39 @@ def get_index(self, url_type, /):
...
+            )
+            for key, value in previous.items():
+                if 'thumbnails' == key:
+                    continue
+                if (
+                    _has_index_metadata_value(value) and
+                    not _has_index_metadata_value(response.get(key))
+                ):
+                    response[key] = value
+            update_fields = ['retrieved', 'value']
```
tcely: The save implementation uses `frozenset`, so I tend to use a `set` instead of a `list` for this.

tubesync/sync/youtube.py
```diff
@@ -81,10 +81,69 @@ def get_channel_id(url):
...
+    thumbnail_id = thumbnail.get('id')
+    if thumbnail_id in ('avatar_uncropped', 'banner_uncropped'):
+        return thumbnail_id,
+
+    def dimension(name):
+        value = thumbnail.get(name)
+        try:
+            return int(value)
+        except (TypeError, ValueError):
+            return value
```
tcely: Wouldn't it be better to guarantee an `int` here?

tubesync/sync/models/source.py
```diff
@@ -627,18 +587,49 @@ def get_index(self, url_type, /):
...
         else:
             if not isinstance(response, dict):
                 return entries
-            # Keep the channel image URLs from this response so downloading
-            # source images does not need a separate extraction later.
-            self.save_indexed_image_urls(response)
-            entries = response.get('entries', list())
+            response = response.copy()
+            entries = response.pop('entries', list())
+            site = response.get('extractor_key')
```
tcely:
```suggestion
            default_site = self.videos.model._meta.get_field('site').get_default()
            site = response.get('extractor_key', default_site)
```

Using the site default in the get call makes more sense to me.

tcely (changes requested):

tubesync/sync/models/source.py
ryanduguid: Removed the copy. `get_index` now consumes `entries` before storing the response, and the focused test verifies that the response no longer retains `entries`.

tubesync/sync/models/source.py
ryanduguid: Simplified this to `bool(value) or value in (0, False,)`. The values originate from JSON metadata, so this keeps the intended handling of non-empty values plus numeric zero and false.

tubesync/sync/models/source.py
ryanduguid: Changed update_fields to a set and add site with set.add, matching Django's set-based handling.

tubesync/sync/youtube.py
ryanduguid: Dimension normalisation now returns int(0) for missing or invalid values. A focused test passes an unhashable JSON dimension and verifies thumbnail merging deduplicates without error.

tubesync/sync/models/source.py
ryanduguid: Kept site outside the lookup deliberately. Including the default site in get_or_create would create a second row when an incomplete response first uses the default and a later response supplies YoutubeTab. The row is keyed by source plus index URL; the later site updates that same row. Added a test covering this sequence.

tcely: Your comment didn't make sense regarding my suggestion, so I added an example to my comment.

You can update `get_or_create` to use `defaults=dict(site=site),` also.

tubesync/sync/youtube.py
```diff
@@ -81,10 +81,69 @@ def get_channel_id(url):
...
     thumbnail_url = None
+    if not isinstance(response, dict):
+        return avatar_url, banner_url, thumbnail_url
+    max_height = 0
+    for thumbnail in _thumbnail_items(response.get('thumbnails')):
+        if not isinstance(thumbnail, dict):
+            continue
+        thumbnail_height = thumbnail.get('height')
+        try:
+            thumbnail_height = int(thumbnail_height)
```
tcely: When you are guaranteed an `int` earlier, this entire `try` becomes unnecessary.

tubesync/sync/youtube.py
```diff
@@ -81,10 +81,69 @@ def get_channel_id(url):
...
     avatar_url = None
     banner_url = None
     thumbnail_url = None
+    if not isinstance(response, dict):
+        return avatar_url, banner_url, thumbnail_url
+    max_height = 0
+    for thumbnail in _thumbnail_items(response.get('thumbnails')):
+        if not isinstance(thumbnail, dict):
+            continue
+        thumbnail_height = thumbnail.get('height')
```
tcely:
```suggestion
        thumbnail_height = thumbnail.get('height', int())
```

tcely (changes requested):

tcely:
### Lint findings

- [x] sync/models/source.py

```
sync/models/source.py:
   55:12 RUF012 Mutable default value for class attribute
   65:18 RUF012 Mutable default value for class attribute
  571:16 BLE001 Do not catch blind exception: `Exception`
  582:13 TRY004 Prefer `TypeError` exception for invalid type
  582:19 TRY002 Create your own exception
  642:9  SIM102 Use a single `if` statement instead of nested `if` statements
```

<details><summary>Full</summary>

```
RUF012 Mutable default value for class attribute
  --> sync/models/source.py:55:12
   |
54 |       # Format to use to display a URL for the source
55 |       URLS = dict(zip(
   |  ____________^
56 | |         YouTube_SourceType.values,
57 | |         (
58 | |             'https://www.youtube.com/c/{key}',
59 | |             'https://www.youtube.com/channel/{key}',
60 | |             'https://www.youtube.com/playlist?list={key}',
61 | |         ),
62 | |     ))
   | |______^
63 |
64 |       # Format used to create indexable URLs
   |
help: Consider initializing in `__init__` or annotating with `typing.ClassVar`

RUF012 Mutable default value for class attribute
  --> sync/models/source.py:65:18
   |
64 |       # Format used to create indexable URLs
65 |       INDEX_URLS = dict(zip(
   |  __________________^
66 | |         YouTube_SourceType.values,
67 | |         (
68 | |             'https://www.youtube.com/c/{key}/{type}',
69 | |             'https://www.youtube.com/channel/{key}/{type}',
70 | |             'https://www.youtube.com/playlist?list={key}',
71 | |         ),
72 | |     ))
   | |______^
73 |
74 |       # Callback functions to get a list of media from the source
   |
help: Consider initializing in `__init__` or annotating with `typing.ClassVar`

BLE001 Do not catch blind exception: `Exception`
   --> sync/models/source.py:571:16
    |
569 |         try:
570 |             return self.media_format.format(**self.example_media_format_dict)
571 |         except Exception:
    |                ^^^^^^^^^
572 |             return ''
    |

TRY004 Prefer `TypeError` exception for invalid type
   --> sync/models/source.py:582:13
    |
580 |         indexer = self.INDEXERS.get(self.source_type, None)
581 |         if not callable(indexer):
582 |             raise Exception(f'Source type f"{self.source_type}" has no indexer')
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
583 |         days = None
584 |         if self.download_cap_date:
    |

TRY002 Create your own exception
   --> sync/models/source.py:582:19
    |
580 |         indexer = self.INDEXERS.get(self.source_type, None)
581 |         if not callable(indexer):
582 |             raise Exception(f'Source type f"{self.source_type}" has no indexer')
    |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
583 |         days = None
584 |         if self.download_cap_date:
    |

SIM102 Use a single `if` statement instead of nested `if` statements
   --> sync/models/source.py:642:9
    |
641 |           # Playlists do something different that I have yet to figure out
642 | /         if not self.is_playlist:
643 | |             if self.index_streams:
    | |__________________________________^
644 |                   streams = self.get_index('streams')
645 |                   if entries.maxlen is None or 0 == len(entries):
    |
help: Combine `if` statements using `and`
```

</details>

- [x] sync/youtube.py

```
sync/youtube.py:
   46:5  PIE790 [*] Unnecessary `pass` statement
  181:12 SIM118 Use `key not in dict` instead of `key not in dict.keys()`
  207:61 SIM118 Use `key in dict` instead of `key in dict.keys()`
  244:9  SIM118 Use `key in dict` instead of `key in dict.keys()`
  266:20 SIM210 Use `bool(...)` instead of `True if ... else False`
  398:18 SIM211 Use `not ...` instead of `False if ... else True`
  399:20 SIM210 Use `bool(...)` instead of `True if ... else False`
```

<details><summary>Full</summary>

```
PIE790 [*] Unnecessary `pass` statement
  --> sync/youtube.py:46:5
   |
44 |         Generic wrapped error for all errors that could be raised by youtube-dl.
45 |     '''
46 |     pass
   |     ^^^^
help: Remove unnecessary `pass`
   |
45 |     '''
   -     pass
46 |
   |

SIM118 Use `key not in dict` instead of `key not in dict.keys()`
   --> sync/youtube.py:181:12
    |
179 |             raise TypeError(f'response must be a dict, got "{type(response)}" instead')
180 |
181 |         if 'availability' not in response.keys():
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
182 |             return False
    |
help: Remove `.keys()`

SIM118 Use `key in dict` instead of `key in dict.keys()`
   --> sync/youtube.py:207:61
    |
205 |     default_opts = yt_dlp.parse_options([]).options
206 |     class NoDefaultValue: pass # a unique Singleton, that may be checked for later
207 |     user_set = lambda k, d, default=NoDefaultValue: d[k] if k in d.keys() else default
    |                                                             ^^^^^^^^^^^^^
208 |     default_paths = user_set('paths', default_opts.__dict__, dict())
209 |     paths = user_set('paths', opts, default_paths)
    |
help: Remove `.keys()`

SIM118 Use `key in dict` instead of `key in dict.keys()`
   --> sync/youtube.py:244:9
    |
242 |         pl_infojson=f'{cache_directory_path}/infojson/playlist/{playlist_infojson}.%(ext)s',
243 |     )
244 |     for k in OUTTMPL_TYPES.keys():
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^
245 |         outtmpl.setdefault(k, '')
246 |     sleep_interval_requests = getattr(settings, 'YOUTUBE_INFO_SLEEP_REQUESTS', 1)
    |
help: Remove `.keys()`

SIM210 Use `bool(...)` instead of `True if ... else False`
   --> sync/youtube.py:266:20
    |
264 |         'skip_unavailable_fragments': False,
265 |         'sleep_interval_requests': sleep_interval_requests,
266 |         'verbose': True if settings.DEBUG else False,
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
267 |         'writeinfojson': True,
268 |     })
    |
help: Replace with `bool(...)

SIM211 Use `not ...` instead of `False if ... else True`
   --> sync/youtube.py:398:18
    |
396 |         'outtmpl': os.path.basename(output_file),
397 |         'remuxvideo': pp_opts.remuxvideo,
398 |         'quiet': False if settings.DEBUG else True,
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
399 |         'verbose': True if settings.DEBUG else False,
400 |         'noprogress': None if settings.DEBUG else True,
    |
help: Replace with `not ...`

SIM210 Use `bool(...)` instead of `True if ... else False`
   --> sync/youtube.py:399:20
    |
397 |         'remuxvideo': pp_opts.remuxvideo,
398 |         'quiet': False if settings.DEBUG else True,
399 |         'verbose': True if settings.DEBUG else False,
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
400 |         'noprogress': None if settings.DEBUG else True,
401 |         'writeinfojson': info_json,
    |
help: Replace with `bool(...)
```

</details>

tubesync/sync/models/source.py
```diff
@@ -485,6 +492,19 @@ def make_directory(self):
     def get_image_url(self):
         return get_youtube_image_info(self.url)
 
+    def get_image_urls(self, qs):
+        avatar_url = banner_url = thumbnail_url = None
+        for metadata in qs.order_by('-retrieved', '-created'):
```
tcely: Does this prefer `videos` over `streams` for sources with both enabled?

tubesync/sync/youtube.py
```diff
@@ -81,10 +84,69 @@ def get_channel_id(url):
             else:
                 return channel_id
 
-def get_image_info(url):
+def _thumbnail_items(value):
+    if not isinstance(value, (list, tuple)):
+        return tuple()
+    return value
```
tcely:
```suggestion
    return value if isinstance(value, (list, tuple)) else tuple()
```

tubesync/sync/models/source.py
ryanduguid: The current head supplies site only through get_or_create defaults. It remains outside the filter/get lookup, so the source plus index-URL row stays stable; the existing incomplete-default-site then YoutubeTab test verifies one row is retained and its site is updated.

tubesync/sync/youtube.py
ryanduguid: Consolidated dimension conversion in _thumbnail_dimension at 7f4f7d80. get_image_urls now receives a guaranteed integer from that helper and no longer has its own try block; malformed present values still normalise to zero.

tubesync/sync/youtube.py
ryanduguid: _thumbnail_dimension now uses thumbnail.get(name, int()) before conversion at 7f4f7d80. Missing dimensions therefore default to zero, while invalid present values are also retained safely as zero.

tubesync/sync/models/source.py
ryanduguid: It did not previously: the newest retrieved row won regardless of index type. Fixed at 7f4f7d80 so videos metadata is processed first, with streams metadata only filling missing image types. The regression test first failed with the streams avatar and thumbnail, then passed with videos preferred.

tubesync/sync/youtube.py
ryanduguid: Applied the conditional-expression form at 7f4f7d80 with unchanged list/tuple handling.

tcely (approved):
Thanks again!

meeb (approved):

PR #1569, opened 2026-08-27
fix: truncate generated filenames to the filesystem byte limit

### LINKED ISSUE #522: [bug] can't download files which filename will exceed 255 characters
From my added sources, there are 14 tasks stuck, because they fail to download:
`ERROR: unable to open for writing: [Errno 36] File name too long: '/downloads/video/Kochou Cocoa/2021-07-18 - 【3Dio#ASMR 】同接１５０超えるまでやめれません！耳はむ多めで甘々でデレデレになっちゃう最高の癒し♡たくさん甘えて？♡耳フー♡ please Request【#胡蝶ここあ Vtuber】 (DdMVc7-Ks8A).f299.mp4.part'`
The original video has a character length of 98 characters but because they are unicode, this translates to 225 ASCII characters. After tubesync adds the pre- and suffix as well as the part file extensions it easily exceeds 255 characters.

There needs to be a check of the the maximum resulting character limit (when there are more parts to fetch the resulting filename gradually grows) and the Video title needs to be trimmed.

As a workaround I managed to download the videos separately and added them as external videos.

Agi-Asi:
### Problem

Filesystems limit each path component to **255 bytes, not characters**. Multi-byte titles (CJK, emoji) hit that with far fewer characters — the reporter's 98-character Japanese title is already 225+ bytes before TubeSync's prefixes and yt-dlp's `.fNNN.mp4.part(-FragNN)` suffixes are appended — and the download then fails permanently:

```
ERROR: unable to open for writing: [Errno 36] File name too long
```

The task retries forever with no way out short of changing the source's media format (#522).

### Fix

`Media.filename` now clamps the final name component to a **200-byte budget** (leaving headroom for the suffixes yt-dlp appends during download, per the direction in the issue thread — "we only need to ensure our final format based filename is not over the limit"):

- Only the name component is shortened; directories from the format string are untouched.
- Bytes are removed from the **middle of the stem**, keeping its start (the beginning of the title) and its end (the unique `{key}`/`{format}` suffixes of the default media format), joined by `_..._` — so uniqueness of generated names survives truncation.
- Truncation never splits a multi-byte character (partial sequences are dropped when decoding).
- A warning is logged whenever a name is shortened.

Names within the limit are returned **byte-identical**, so existing installs see no change unless they were already broken.

### Testing

- New tests: a 360-byte multi-byte title is clamped under the budget with directories, key suffix, and UTF-8 validity preserved; names within the limit are unchanged.
- Full Django suite: `manage.py test` — 53 tests, OK.
- The CI ruff gate (`--select 'C4,E4,E7,E9,F'` with the workflow's ignore list) passes.

Fixes #522


tubesync/sync/models/media.py
```diff
@@ -841,7 +842,22 @@ def filename(self):
...
+        result = '.' + result if '/' == result[0] else result
+        # Filesystems limit each path component to 255 bytes (not
+        # characters), and multi-byte titles can blow past that with far
+        # fewer characters — downloads then fail with:
+        #   [Errno 36] File name too long
+        # (issue #522). Only the final component (the name) is shortened;
+        # any directories in the format string are preserved. The budget
+        # leaves headroom for suffixes appended during download
+        # (`.fNNN.ext.part-FragNNN.part` and thumbnail/subtitle siblings).
+        head, _, tail = result.rpartition('/')
```
tcely: It would make more sense to use pathlib.Path to parse the string for us.

tubesync/sync/tests/test_filepath.py
```diff
@@ -158,6 +158,34 @@ def test_media_filename(self):
                          ('no-fancy-stuff-title_test_720p-720x1280-opus'
                           '-vp9-30fps-hdr.mkv'))
 
+    def test_media_filename_truncates_to_filesystem_byte_limit(self):
+        # Filesystems limit each name component to 255 bytes (issue #522).
+        # Multi-byte titles reach that with far fewer characters, and the
+        # download then fails with '[Errno 36] File name too long'.
+        import json
```
tcely: This belongs at the top of the file.

tubesync/sync/tests/test_filepath.py
```diff
@@ -158,6 +158,34 @@ def test_media_filename(self):
...
+        filename = long_title_media.filename
+        directory, _, name = filename.rpartition('/')
+        # Directories from the format string survive untouched
+        self.assertEqual('2017', directory)
+        # The name component fits in the byte budget...
+        self.assertLessEqual(len(name.encode('utf-8')), 200)
+        # ... keeps its extension and key suffix material intact ...
+        self.assertTrue(name.endswith('_longkey.mkv'))
+        # ... and was not cut mid multi-byte character (encodes cleanly)
+        name.encode('utf-8').decode('utf-8')
```
tcely, line 184: Also test bytes that result in encoding/decoding errors.

tcely (changes requested):
This looks fairly good.

I dislike using string functions for parsing paths instead of `pathlib.Path`.

Also, we should test the code against bytes known to cause encoding / decoding issues. We need filenames to be workable at all times.

--------

### Lint findings

- [x] tubesync/common/utils.py
```
common/utils.py:
  223:5  RUF059 Unpacked variable `user_pass` is never used
  247:5  FURB188 [*] Prefer `str.removeprefix()` over conditionally replacing with slice.
  289:9  TRY004 Prefer `TypeError` exception for invalid type
  304:9  TRY004 Prefer `TypeError` exception for invalid type
  325:9  TRY004 Prefer `TypeError` exception for invalid type
  354:12 UP032 [*] Use f-string instead of `format` call
```

<details><summary>full format</summary>

```
RUF059 Unpacked variable `user_pass` is never used
   --> common/utils.py:223:5
    |
221 |         raise DatabaseConnectionError('Database connection string netloc must be in '
222 |                                       'the format of user:pass@host')
223 |     user_pass, host_port = host_parts
    |     ^^^^^^^^^
224 |     username, password = user_pass_parts
225 |     host_port_parts = host_port.split(':')
    |
help: Prefix it with an underscore or any other dummy variable pattern

FURB188 [*] Prefer `str.removeprefix()` over conditionally replacing with slice.
   --> common/utils.py:247:5
    |
245 |           raise DatabaseConnectionError('Database connection host must be a hostname or '
246 |                                         'a hostname:port combination')
247 | /     if database.startswith('/'):
248 | |         database = database[1:]
    | |_______________________________^
249 |       if not database:
250 |           raise DatabaseConnectionError('Database connection string path must be a '
    |
help: Use removeprefix instead of assignment conditional upon startswith.
    |
246 |                                       'a hostname:port combination')
    -     if database.startswith('/'):
    -         database = database[1:]
247 +     database = database.removeprefix('/')
248 |     if not database:
    |

TRY004 Prefer `TypeError` exception for invalid type
   --> common/utils.py:289:9
    |
287 | def clean_filename(filename):
288 |     if not isinstance(filename, str):
289 |         raise ValueError(f'filename must be a str, got {type(filename)}')
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
290 |     to_scrub = r'<>\/:*?"|%'
291 |     for char in list(to_scrub):
    |

TRY004 Prefer `TypeError` exception for invalid type
   --> common/utils.py:304:9
    |
302 | def clean_emoji(s):
303 |     if not isinstance(s, str):
304 |         raise ValueError(f'parameter must be a str, got {type(s)}')
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
305 |     return emoji.replace_emoji(s)
    |

TRY004 Prefer `TypeError` exception for invalid type
   --> common/utils.py:325:9
    |
323 |     '''
324 |     if not isinstance(filename, str):
325 |         raise ValueError(f'filename must be a str, got {type(filename)}')
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
326 |     if len(filename.encode(encoding)) <= max_bytes:
327 |         return filename
    |

UP032 [*] Use f-string instead of `format` call
   --> common/utils.py:354:12
    |
352 |     minutes = seconds // 60
353 |     seconds %= 60
354 |     return '{:02d}:{:02d}:{:02d}'.format(hour, minutes, seconds)
    |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
help: Convert to f-string
    |
353 |     seconds %= 60
    -     return '{:02d}:{:02d}:{:02d}'.format(hour, minutes, seconds)
354 +     return f'{hour:02d}:{minutes:02d}:{seconds:02d}'
355 |
    |
```

</details>

- [x] tubesync/sync/models/media.py
```
sync/models/media.py:
    67:23 RUF012 Mutable default value for class attribute
    84:19 RUF012 Mutable default value for class attribute
   281:9  B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   310:13 B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   378:13 RUF059 Unpacked variable `combined_match` is never used
   382:17 RUF059 Unpacked variable `audio_match` is never used
   383:17 RUF059 Unpacked variable `video_match` is never used
   595:9  B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   600:39 B006 Do not use mutable data structures for argument defaults
   624:13 RUF059 Unpacked variable `created` is never used
   631:9  B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   668:16 SIM118 Use `key in dict` instead of `key in dict.keys()`
   680:16 BLE001 Do not catch blind exception: `Exception`
   719:13 B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   721:16 BLE001 Do not catch blind exception: `Exception`
   743:13 PIE790 [*] Unnecessary `pass` statement
   780:20 DTZ007 Naive datetime constructed using `datetime.datetime.strptime()` without %z
   783:13 PIE790 [*] Unnecessary `pass` statement
   875:17 RUF059 Unpacked variable `ext` is never used
  1080:13 TRY004 Prefer `TypeError` exception for invalid type
  1080:19 TRY002 Create your own exception
  1085:13 SIM118 Use `key not in dict` instead of `key not in dict.keys()`
  1196:26 RUF059 Unpacked variable `fuzzy_prefix_path` is never used
```

<details><summary>full format</summary>

```
RUF012 Mutable default value for class attribute
  --> sync/models/media.py:67:23
   |
65 |       # Maps standardised names to names used in source metdata
66 |       _same_name = lambda n, k=None: {k or n: _srctype_dict(n) }
67 |       METADATA_FIELDS = {
   |  _______________________^
68 | |         **(_same_name('upload_date')),
69 | |         **(_same_name('timestamp')),
70 | |         **(_same_name('title')),
71 | |         **(_same_name('fulltitle')),
72 | |         **(_same_name('description')),
73 | |         **(_same_name('duration')),
74 | |         **(_same_name('formats')),
75 | |         **(_same_name('categories')),
76 | |         **(_same_name('average_rating', 'rating')),
77 | |         **(_same_name('age_limit')),
78 | |         **(_same_name('uploader')),
79 | |         **(_same_name('like_count', 'upvotes')),
80 | |         **(_same_name('dislike_count', 'downvotes')),
81 | |         **(_same_name('playlist_title')),
82 | |     }
   | |_____^
83 |
84 |       STATE_ICONS = dict(zip(
   |
help: Consider initializing in `__init__` or annotating with `typing.ClassVar`

RUF012 Mutable default value for class attribute
  --> sync/models/media.py:84:19
   |
82 |       }
83 |
84 |       STATE_ICONS = dict(zip(
   |  ___________________^
85 | |         MediaState.values,
86 | |         (
87 | |             '<i class="far fa-question-circle" title="Unknown download state"></i>',
88 | |             '<i class="far fa-clock" title="Scheduled to download"></i>',
89 | |             '<i class="fas fa-download" title="Downloading now"></i>',
90 | |             '<i class="far fa-check-circle" title="Downloaded"></i>',
91 | |             '<i class="fas fa-exclamation-circle" title="Skipped"></i>',
92 | |             '<i class="fas fa-stop-circle" title="Media downloading disabled at source"></i>',
93 | |             '<i class="fas fa-exclamation-triangle" title="Error downloading"></i>',
94 | |         )
95 | |     ))
   | |______^
96 |
97 |       uuid = models.UUIDField(
   |
help: Consider initializing in `__init__` or annotating with `typing.ClassVar`

B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   --> sync/models/media.py:281:9
    |
280 |     def save(self, force_insert=False, force_update=False, using=None, update_fields=None):
281 |         setattr(self, '_cached_metadata_dict', None)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
282 |         # Correct the path after a source is renamed
283 |         if self.created and self.downloaded and not self.media_file_exists:
    |
help: Replace `setattr` with assignment
    |
280 |     def save(self, force_insert=False, force_update=False, using=None, update_fields=None):
    -         setattr(self, '_cached_metadata_dict', None)
281 +         self._cached_metadata_dict = None
282 |         # Correct the path after a source is renamed
    |

B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   --> sync/models/media.py:310:13
    |
308 |             self.title = self.metadata_title[:200] or self.title
309 |             self.duration = self.metadata_duration or self.duration
310 |             setattr(self, '_cached_metadata_dict', None)
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
311 |             if update_fields is not None:
312 |                 # If only some fields are being updated, make sure we update title and duration if metadata changes
    |
help: Replace `setattr` with assignment
    |
309 |             self.duration = self.metadata_duration or self.duration
    -             setattr(self, '_cached_metadata_dict', None)
310 +             self._cached_metadata_dict = None
311 |             if update_fields is not None:
    |

RUF059 Unpacked variable `combined_match` is never used
   --> sync/models/media.py:378:13
    |
376 |                 return False
377 |         else:
378 |             combined_match, combined_format = self.get_best_combined_format()
    |             ^^^^^^^^^^^^^^
379 |             if combined_format:
380 |                 return str(combined_format)
    |
help: Prefix it with an underscore or any other dummy variable pattern

RUF059 Unpacked variable `audio_match` is never used
   --> sync/models/media.py:382:17
    |
380 |                 return str(combined_format)
381 |             else:
382 |                 audio_match, audio_format = self.get_best_audio_format()
    |                 ^^^^^^^^^^^
383 |                 video_match, video_format = self.get_best_video_format()
384 |                 if audio_format and video_format:
    |
help: Prefix it with an underscore or any other dummy variable pattern

RUF059 Unpacked variable `video_match` is never used
   --> sync/models/media.py:383:17
    |
381 |             else:
382 |                 audio_match, audio_format = self.get_best_audio_format()
383 |                 video_match, video_format = self.get_best_video_format()
    |                 ^^^^^^^^^^^
384 |                 if audio_format and video_format:
385 |                     return f'{video_format}+{audio_format}'
    |
help: Prefix it with an underscore or any other dummy variable pattern

B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   --> sync/models/media.py:595:9
    |
593 |     def metadata_clear(self, /, *, save=False):
594 |         self.metadata = None
595 |         setattr(self, '_cached_metadata_dict', None)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
596 |         if save:
597 |             self.save()
    |
help: Replace `setattr` with assignment
    |
594 |         self.metadata = None
    -         setattr(self, '_cached_metadata_dict', None)
595 +         self._cached_metadata_dict = None
596 |         if save:
    |

B006 Do not use mutable data structures for argument defaults
   --> sync/models/media.py:600:39
    |
600 |     def metadata_dumps(self, arg_dict=dict()):
    |                                       ^^^^^^
601 |         fallback = dict()
602 |         try:
    |
help: Replace with `None`; initialize within function

RUF059 Unpacked variable `created` is never used
   --> sync/models/media.py:624:13
    |
622 |         )
623 |         md_model = self._meta.fields_map.get('new_metadata').related_model
624 |         md, created = md_model.objects.filter(
    |             ^^^^^^^
625 |             source__isnull=True,
626 |         ).get_or_create(
    |
help: Prefix it with an underscore or any other dummy variable pattern

B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   --> sync/models/media.py:631:9
    |
629 |             key=self.key,
630 |         )
631 |         setattr(self, '_cached_metadata_dict', None)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
632 |         return md.ingest_metadata(data)
    |
help: Replace `setattr` with assignment
    |
630 |         )
    -         setattr(self, '_cached_metadata_dict', None)
631 +         self._cached_metadata_dict = None
632 |         return md.ingest_metadata(data)
    |

SIM118 Use `key in dict` instead of `key in dict.keys()`
   --> sync/models/media.py:668:16
    |
666 |                     using_table = True
667 |                     old_mdl = len(str(data))
668 |             if '_reduce_data_ran_at' in data.keys():
    |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
669 |                 total_seconds = data['_reduce_data_ran_at']
670 |                 assert isinstance(total_seconds, int), type(total_seconds)
    |
help: Remove `.keys()`

BLE001 Do not catch blind exception: `Exception`
   --> sync/models/media.py:680:16
    |
678 |             filtered_data['_reduce_data_ran_at'] = round((now - self.posix_epoch).total_seconds())
679 |             filtered_json = self.metadata_dumps(arg_dict=filtered_data)
680 |         except Exception as e:
    |                ^^^^^^^^^
681 |             log.exception('reduce_data: %s', e)
682 |         else:
    |

B010 [*] Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
   --> sync/models/media.py:719:13
    |
717 |             except ObjectDoesNotExist:
718 |                 pass
719 |             setattr(self, '_cached_metadata_dict', data)
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
720 |             return data
721 |         except Exception:
    |
help: Replace `setattr` with assignment
    |
718 |                 pass
    -             setattr(self, '_cached_metadata_dict', data)
719 +             self._cached_metadata_dict = data
720 |             return data
    |

BLE001 Do not catch blind exception: `Exception`
   --> sync/models/media.py:721:16
    |
719 |             setattr(self, '_cached_metadata_dict', data)
720 |             return data
721 |         except Exception:
    |                ^^^^^^^^^
722 |             return {}
    |

PIE790 [*] Unnecessary `pass` statement
   --> sync/models/media.py:743:13
    |
741 |         except (TypeError, ValueError,) as e:
742 |             log.warn(f'Could not compute published from timestamp for: {self.source} / {self} with "{e}"')
743 |             pass
    |             ^^^^
744 |         else:
745 |             return self.posix_epoch + timedelta(seconds=timestamp_float)
    |
help: Remove unnecessary `pass`
    |
742 |             log.warn(f'Could not compute published from timestamp for: {self.source} / {self} with "{e}"')
    -             pass
743 |         else:
    |

DTZ007 Naive datetime constructed using `datetime.datetime.strptime()` without %z
   --> sync/models/media.py:780:20
    |
778 |             return None
779 |         try:
780 |             return datetime.strptime(upload_date_str, '%Y%m%d')
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
781 |         except (AttributeError, ValueError) as e:
782 |             log.debug(f'Media.upload_date: {self.source} / {self}: strptime: {e}')
    |
help: Call `.replace(tzinfo=<timezone>)` or `.astimezone()` to convert to an aware datetime

PIE790 [*] Unnecessary `pass` statement
   --> sync/models/media.py:783:13
    |
781 |         except (AttributeError, ValueError) as e:
782 |             log.debug(f'Media.upload_date: {self.source} / {self}: strptime: {e}')
783 |             pass
    |             ^^^^
784 |         return None
    |
help: Remove unnecessary `pass`
    |
782 |             log.debug(f'Media.upload_date: {self.source} / {self}: strptime: {e}')
    -             pass
783 |         return None
    |

RUF059 Unpacked variable `ext` is never used
   --> sync/models/media.py:875:17
    |
873 |         # So, we do not care about the different directories
874 |         # used for filename in the cases above.
875 |         prefix, ext = os.path.splitext(os.path.basename(filename))
    |                 ^^^
876 |         return prefix
    |
help: Prefix it with an underscore or any other dummy variable pattern

TRY004 Prefer `TypeError` exception for invalid type
    --> sync/models/media.py:1080:13
     |
1078 |           indexer = self.INDEXERS.get(self.source.source_type, None)
1079 |           if not callable(indexer):
1080 | /             raise Exception(f'Media with source type f"{self.source.source_type}" '
1081 | |                             f'has no indexer')
     | |______________________________________________^
1082 |           response = indexer(self.url)
1083 |           no_formats_available = (
     |

TRY002 Create your own exception
    --> sync/models/media.py:1080:19
     |
1078 |           indexer = self.INDEXERS.get(self.source.source_type, None)
1079 |           if not callable(indexer):
1080 |               raise Exception(f'Media with source type f"{self.source.source_type}" '
     |  ___________________^
1081 | |                             f'has no indexer')
     | |______________________________________________^
1082 |           response = indexer(self.url)
1083 |           no_formats_available = (
     |

SIM118 Use `key not in dict` instead of `key not in dict.keys()`
    --> sync/models/media.py:1085:13
     |
1083 |         no_formats_available = (
1084 |             not response or
1085 |             "formats" not in response.keys() or
     |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
1086 |             0 == len(response["formats"])
1087 |         )
     |
help: Remove `.keys()`

RUF059 Unpacked variable `fuzzy_prefix_path` is never used
    --> sync/models/media.py:1196:26
     |
1194 |                         if not fuzzy_path.exists():
1195 |                             continue
1196 |                         (fuzzy_prefix_path, fuzzy_stem) = directory_and_stem(fuzzy_path, True)
     |                          ^^^^^^^^^^^^^^^^^
1197 |                         old_file_str = fuzzy_path.name
1198 |                         new_file_str = new_stem + old_file_str[len(fuzzy_stem):]
     |
help: Prefix it with an underscore or any other dummy variable pattern
```

</details>

Agi-Asi:
Thanks for the review (and the lint fixups pushed to the branch — much appreciated). Addressed all three points:

- **pathlib instead of string parsing**: `truncate_filename_bytes` now uses `PurePosixPath` for stem/suffix, and `Media.filename` uses `PurePosixPath.with_name()` instead of `rpartition('/')`.
- **import at top**: inherited via your commits; the new `PurePosixPath` import rides the existing `pathlib` import line in both files.
- **encoding/decoding stress tests**: added a parametrized test over bytes known to cause trouble — cuts landing mid-4-byte-emoji, combining sequences (e+U+0301), ZWJ family emoji, mixed 1/3-byte widths at every boundary parity, RTL text, dotfiles, no-extension, and oversized-extension names — asserting the byte budget, valid UTF-8 round-trip, and non-empty output for each. Also switched non-str input to `TypeError`.

Full suite: 55 tests OK; the workflow's ruff gate passes.

tcely (changes requested):
To implement the idea from the related issue about using a short known clean unique filename while the download is happening, then renaming the files later, I am considering modifying the task that downloads the media file.

By creating a copy of the media and source inside that task we can provide any format that we like then attach the actual file on disk to the original media instance and update the appropriate attributes.

Since @Agi-Asi didn't respond, I think we will defer those changes for later.

FaySmash:
I would propose not sacrificing 5 bytes `_..._` to join the filename segments but use the 2 byte symbol `…` or 1 byte symbol `~` instead.
Also, maybe expose it as user parameter if text should get truncated from the front, middle or end depending on the source (channels have varying naming schemes).

tcely (approved):

meeb (approved):

PR #1570, opened 2026-08-27
feat: bulk admin actions for `Media.skip`

### LINKED ISSUE #493: FEATURE REQUEST: Bulk editing in the Django admin control panel
Hi there - it doesn't seem like it would be hard to do  - perhaps you could start initially with key features, but would it be possible to add bulk edit in for at the Django admin interface for some key functions on some objects, such as Media?  A real time saver would be to expose actions such as enabling/disabling Skip or Can download.  This would be extremely helpful when editing a large number of objects instead of having to do each individually through the interface.   Thank you for your work on this.

image
image

Agi-Asi:
### Problem

Managing a large library through the Django admin requires editing every media item individually to toggle **Skip** or **Can download** — the exact pain point in #493, where bulk-editing those two flags was requested ("A real time saver would be to expose actions such as enabling/disabling Skip or Can download").

### Change

Adds the standard Django admin bulk actions to `MediaAdmin`, scoped to exactly the two fields requested:

- **Set / unset "skip"** — also sets `manual_skip` to match, so a bulk un-skip isn't immediately re-skipped by the filtering logic, and a bulk skip is recorded as the operator's explicit decision.
- **Set / unset "can download"**.

Also adds `list_filter` for `can_download` / `skip` / `downloaded`, so the rows to bulk-edit can be narrowed down first (without a filter the actions are much less useful on big libraries).

Implementation uses `queryset.update()`, so a bulk action across thousands of rows is a single `UPDATE` statement rather than thousands of model saves.

### Testing

- New `sync/tests/test_admin.py` covers both action pairs end-to-end against real model rows (including the `manual_skip` coupling) plus action registration.
- Full Django suite: `manage.py test` — 54 tests, OK.
- The CI ruff gate passes.

Fixes #493


meeb (commented):
This seems fine with to me. @tcely is toggling flags sufficient to eventually trigger the required worker events?

tcely:
> Is toggling flags sufficient to eventually trigger the required worker events?

I don't understand the question.

meeb:
As in, is just setting `can_download` to `True` sufficient to trigger a task at a later date, or does this need to have `.save()` called on each object as well to re-evaluate the state of the media item.

tcely:
The `media_post_save` signal is the only place the `can_download` attribute is recalculated.

That is eventually triggered for every `Media` instance per source: A) after editing a source or B) after indexing a source.

Is this sufficient, or do you believe a `save_media` task needs to be added to the tasks queue?

Agi-Asi:
Good question re: signals — since `queryset.update()` deliberately bypasses `save()`/`media_post_save`, the flags flip immediately but re-evaluation waits for the next source edit/index pass, as @tcely described. That felt acceptable for an admin bulk tool (it mirrors what the DB-level toggle means), but if you'd prefer eager re-evaluation I'm happy to switch the actions to iterate `queryset` and call `save()` per object (or enqueue a `save_media` task per item) — just tell me which shape you want.

meeb:
I'm fine with it waiting for an Indexing to trigger any events, just checking.

tcely (changes requested):
Please implement a `save_media` task per changed item and add a setting that enables queuing those tasks. By default, the tasks should not be queued.

Agi-Asi:
Done in be4313ba — implemented exactly as requested:

- New setting `SAVE_MEDIA_AFTER_BULK_ACTION` (container env: `TUBESYNC_SAVE_MEDIA_AFTER_BULK_ACTION`), **default `False`** so nothing is queued unless explicitly enabled.
- When enabled, each bulk action queues a `save_media` task per changed item (via `save_media.map()` over the queryset UUIDs), so flags are re-evaluated without waiting for the next source edit/index pass.
- Documented in the README env-var table and `local_settings.py.container`.
- Two new tests: default = `save_media.map` not called; enabled via `override_settings` = called once with exactly the changed UUIDs. All 5 admin tests pass locally, ruff gate clean.

tubesync/sync/tests/test_admin.py
```diff
@@ -0,0 +1,110 @@
...
+        for action in (
+            'enable_skip', 'disable_skip',
+            'enable_can_download', 'disable_can_download',
+        ):
+            self.assertIn(action, self.model_admin.actions)
+
+    def test_save_media_tasks_not_queued_by_default(self):
+        queryset = Media.objects.filter(source=self.source)
+
+        with patch('sync.admin.save_media') as mock_save_media:
```
tcely:
```suggestion
        with patch('sync.tasks.save_media') as mock_save_media:
```

Does patching the import from tasks like this even work??

tubesync/sync/tests/test_admin.py
```diff
@@ -0,0 +1,110 @@
...
+            self.model_admin.enable_skip(self.request, queryset)
+
+        mock_save_media.map.assert_not_called()
+
+    @override_settings(SAVE_MEDIA_AFTER_BULK_ACTION=True)
+    def test_save_media_tasks_queued_when_enabled(self):
+        queryset = Media.objects.filter(source=self.source)
+
+        with patch('sync.admin.save_media') as mock_save_media:
```
tcely:
```suggestion
        with patch('sync.tasks.save_media') as mock_save_media:
```

Again, does mocking the import from `sync.tasks` work?

tubesync/sync/admin.py
```diff
@@ -25,6 +27,50 @@ class MediaAdmin(admin.ModelAdmin):
...
+        'enable_can_download', 'disable_can_download',
+    )
+
+    def _queue_save_media_tasks(self, queryset):
+        '''
+            Optionally queue a `save_media` task per changed item, so flags
+            are re-evaluated without waiting for the next source edit/index.
+            Disabled by default: SAVE_MEDIA_AFTER_BULK_ACTION = True enables it.
+        '''
+        if not getattr(settings, 'SAVE_MEDIA_AFTER_BULK_ACTION', False):
```
tcely:
```suggestion
        if getattr(settings, 'SAVE_MEDIA_AFTER_BULK_ACTION', False):
```

When the setting is `True` we want to queue the tasks.

tubesync/sync/admin.py
```diff
@@ -25,6 +27,50 @@ class MediaAdmin(admin.ModelAdmin):
...
+            return
+        save_media.map(
+            str(media_uuid)
+            for media_uuid in queryset.values_list('uuid', flat=True)
```
tcely: We should probably use `distinct` or something similar to avoid creating duplicate tasks for the same `Media` instance.

Also, `only` for the `uuid` would be a good addition.

README.md
```diff
@@ -267,6 +267,7 @@ useful if you are manually installing TubeSync in some other environment. These
 | TUBESYNC_DIRECTORY_PREFIX    | Enable `video` and `audio` directory prefixes in `/downloads` | True                                                                          |
 | TUBESYNC_SHRINK_NEW          | Filter unneeded information from newly retrieved metadata     | True                                                                          |
 | TUBESYNC_SHRINK_OLD          | Filter unneeded information from metadata loaded from the database | True                                                                     |
+| TUBESYNC_SAVE_MEDIA_AFTER_BULK_ACTION | Queue a `save_media` task per item changed by an admin bulk action, defaults to False | False                                                |
```
tcely:
```suggestion
| TUBESYNC_SAVE_MEDIA_AFTER_BULK_ACTION | Queue a `save_media` task per item changed by an admin bulk action | True                                                |
```

tcely (changes requested):

tubesync/sync/admin.py
tcely:
```suggestion
            for media_uuid in queryset.distinct().values_list('uuid', flat=True).iterator()
```

tcely (approved):

meeb (approved):

PR #1582, opened 2026-09-02
feat(source): add per-source setting to prefer original or default audio

### LINKED ISSUE #1468: Add a setting to choose between original and default audio
> > Do you think it is worth adding a setting to consistently choose either original or default formats?
> 
> Yes that sounds like a nice per-source toggle to have available. It's probably more of a documentation challenge rather than the feature itself to accurately communicate what it does to people.

_Originally posted by @meeb in https://github.com/meeb/tubesync/issues/1452#issuecomment-4293186923_

My first idea was to just use a setting that would apply to all sources.

IshanA2007:
Fixes #1468

## What

Adds `Source.prefer_original_audio`, a per-source toggle (default: on, preserving
current behaviour) controlling whether combined-format matching prefers the
"original" audio track or the "default" one, for videos that offer more than one
audio language.

## Why

#1452 made `get_best_combined_format()` always prefer the `(original)` track over
`(default)`. In that thread @tcely asked whether this should be configurable and
@meeb agreed it'd make a nice per-source toggle, noting the real challenge is
communicating clearly what it does. This implements that toggle.

## How

- New `Source.prefer_original_audio` field + migration
  (`0038_source_prefer_original_audio`), shaped like `prefer_60fps`/`prefer_hdr`.
- `matching.py::get_best_combined_format()`: the "prefer original" branch is now
  gated on `media.source.prefer_original_audio`. When off it falls through to the
  existing `get_fallback_id()` call, which already prefers `(default)` first, so
  there's no new fallback logic.
- Exposed on the add/update source form (`SourceForm.fields`) and the source
  detail page, next to the other "prefer" toggles. `SourceAdmin` doesn't restrict
  `fields`/`fieldsets`, so no admin.py change is needed.
- Added `test_prefer_original_audio_format_matching`, asserting both settings
  against synthetic multi-language combined formats. There was no existing
  coverage of the original/default branch.

## Scope

This mirrors #1452 exactly: it only affects `get_best_combined_format()`.
`get_best_audio_format()` still hardcodes a `(default)` preference and has no
`(original)` handling at all, so the toggle has no effect on sources that
download separate video and audio streams. I kept it that way deliberately:
wiring the toggle into the audio-only path would silently change track selection
for every existing source, which felt out of scope for "add a setting". Happy to
extend it in this PR if you'd prefer the fully consistent behaviour @tcely
described. The help text calls this limitation out so the option can't look
broken to someone using it on a 1080p source.

## Test plan

- `cd tubesync && python manage.py test --verbosity=2` — 61/61 pass (60 before).
- `python manage.py makemigrations sync --check --dry-run` — "No changes detected".
- `ruff check` with the CI invocation — clean.


tubesync/sync/matching.py
```diff
@@ -83,9 +83,10 @@ def get_best_combined_format(media):
     if not matches:
         return False, False
 
-    # prefer original
-    if 'original' in by_fmt_id and 'id' in by_fmt_id['original']:
-        return True, by_fmt_id['original']['id']
+    # prefer original, unless the source prefers the default audio track
+    if media.source.prefer_original_audio:
```
tcely, line 87:
```
   87:5  SIM102 Use a single `if` statement instead of nested `if` statements
```

[collapsible-if (SIM102)](https://docs.astral.sh/ruff/rules/collapsible-if/#collapsible-if-sim102)

tubesync/sync/matching.py
tcely, line 1: ### tubesync/sync/matching.py

```
   25:49 B006 Do not use mutable data structures for argument defaults
   65:9  SIM102 Use a single `if` statement instead of nested `if` statements
   69:9  SIM102 Use a single `if` statement instead of nested `if` statements
   87:5  SIM102 Use a single `if` statement instead of nested `if` statements
  151:9  SIM114 [*] Combine `if` branches using logical `or` operator
  494:13 SIM114 [*] Combine `if` branches using logical `or` operator
  497:13 SIM114 [*] Combine `if` branches using logical `or` operator
```

tubesync/sync/forms.py
tcely, line 1: ### tubesync/sync/forms.py

```
   83:17 INT001 f-string is resolved before function call; consider `_("string %s") % arg`
   87:17 INT001 f-string is resolved before function call; consider `_("string %s") % arg`
  102:17 INT001 f-string is resolved before function call; consider `_("string %s") % arg`
  106:17 INT001 f-string is resolved before function call; consider `_("string %s") % arg`
```

These should all be ignored.
`# ruff: ignore[INT001]`

tubesync/sync/tests/test_frontend.py
tcely, line 1: ### tubesync/sync/tests/test_frontend.py

```
  121:37 PERF102 When using only the values of a dict use the `values()` method
  197:9  FURB188 [*] Prefer `str.removeprefix()` over conditionally replacing with slice.
  259:9  FURB188 [*] Prefer `str.removeprefix()` over conditionally replacing with slice.
  297:9  FURB188 [*] Prefer `str.removeprefix()` over conditionally replacing with slice.
  355:41 DTZ001 `datetime.datetime()` called without a `tzinfo` argument
```

tcely (commented):
Lint findings

tcely:
Superceded by https://github.com/meeb/tubesync/pull/1585

PR #1600, opened 2026-09-21
chore(ci): upgrade test job runner from Ubuntu 22 to 24

tcely:
I don't remember why, but we had `test` pinned to Ubuntu LTS 22.

We don't appear to require that anymore and it will be disappearing in the future.


meeb:
I believe for the ancient version of Django (for the required system Python etc.) required for the old tasks system or something like that. Fine to update.

meeb (approved):

PR #1601, opened 2026-09-21
feat(sqlite): switch journal_mode from TRUNCATE to WAL

jlaska:
## Summary

- Switch SQLite `journal_mode` from `TRUNCATE` to `WAL` in the container
  settings, enabling concurrent readers alongside a single writer
- Add `PRAGMA wal_autocheckpoint = 1000` (SQLite's default, explicit for
  clarity) and `PRAGMA busy_timeout = 5000` to reduce lock contention

## Motivation

Tubesync has significant concurrent write pressure on SQLite — 3+ gunicorn
workers and 4 huey consumer processes (several multi-threaded) all write to
the same database. The codebase already mitigates this with retry decorators,
random delays after saves, and task-level locking. WAL mode reduces this
contention at the database level by allowing readers to proceed without
blocking on writers.

WAL mode also enables compatibility with SQLite replication tools (e.g.
litestream) that require a WAL file to function. Under the current
TRUNCATE/DELETE journal mode, these tools conflict with Django's locking
and cause "database is locked" errors at startup.

Note: Huey's own task-queue SQLite databases already use WAL mode
(`wal_autocheckpoint = 100` in `common/huey.py`).

## Test plan

- [x] Existing Django test suite passes (52/52 non-frontend tests; 13
      pre-existing frontend errors from missing static file symlinks)
- [x] Container starts cleanly with WAL mode active
- [x] `PRAGMA journal_mode;` returns `wal` inside running container
- [x] Migrations run successfully on first attempt without retry logic

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tcely:
Sorry, this won't be merged.

Users are welcome to change their settings. That's why I left the comment about WAL in there.

Your change to `busy_timeout` may interact with the `timeout` setting earlier in unexpected ways too.

jlaska:
@tcely appreciate the quick response!  Is there a mechanism to influence the default behavior when deploying via container image builds?  I don't see a `TUBESYNC_JOURNAL_MODE` (or equivalent env option).

tcely:
@jlaska There sure is.

The local settings file is meant to be included in your `/config` volume.

`/config/tubesync/local_settings.py`

https://github.com/meeb/tubesync/wiki/How-To-Configure#persistent-configuration-docker

PR #1602, opened 2026-09-21
fix: download thumbnails with curl

tcely:
Instead of scheduling multiple tasks to download each URL, use the new `Media.download_thumbnails` function that calls `curl` to download all the URLs in parallel with connection reuse.

If this fails, for whatever reason, the old task still exists and can fetch the missing thumbnail from a single URL.


meeb:
Sounds fine to me. Could we not use pycurl's CurlMulti or similar rather than subprocess spawning curl?

tcely:
> Could we not use pycurl's CurlMulti or similar rather than subprocess spawning curl?

I see absolutely no advantage to reimplementing the `curl` binary in Python. It's probably technically possible, but would be a huge undertaking and is filled with compatibility concerns around various backend and library versions.

meeb:
Doesn't pycurl just use libcurl? What's being re-implemented?

tcely:
The `curl` CLI binary is handling parallel connections and multiplexed reuse alongside protocol negotiation for both TLS and HTTP versions.

Even the proxy support would need to be reimplemented when switching to `pycurl`.

It seems like a huge waste of effort to me.

meeb:
Thanks for the research. I'm not zealot on either approach, I just personally avoid subprocess calling especially in long running worker processes as it has caused no small amount of pain in the past (bugs not killing child processes properly with timeouts, zombie children, worker locks and so on). It's likely better now than when I last encountered problems, admittedly, just an ingrained initial reaction.

You've done the work here so happy to go with whichever version you prefer. If they're generally identical my preference is to avoid the subprocess, however if you feel strongly against it I'm fine with the subprocess spawning.

For logging can set a hook with pycurl, the setopt value for pycurl.DEBUGFUNCTION will fire a callback for extensive logging if needed. A question for another time but it would be interesting to attach a debugger and see why the downloading on one version is slower than another given they're both going to just be calling libcurl to download files in parallel under the hood, their download performance should be identical.

tcely:
I personally prefer not to pull in `pycurl` because we already have a workable CLI command to call.

I did discover that `pycurl` included a slightly newer version of `libcurl`.

The only other advantage that I see for `pycurl` is doing everything entirely in memory.

I'm happy to leave the function there if you want to experiment, but I'll probably move the import inside it and go with the subprocess version.

meeb (approved):

PR #1604, opened 2026-09-23
fix: improve source deletion

### LINKED ISSUE #1603: Deleting source ends in SQL error
I've been cleaning up my sources, simplifying my setup, and several sources fail to delete, returning the following error:
```
Deleting all media for source "source-name"
Error: "a foreign key constraint fails (`tubesync`.`sync_media`, CONSTRAINT `sync_media_source_id_36827e1d_fk_sync_source_uuid` FOREIGN KEY (`source_id`) REFERENCES `sync_source` (`uuid`))')"
```
I only manage my downloaded videos through TubeSync, so TubeSync is in charge of deleting files and cleaning up the database-entries for those videos. Is there something causing a race-condition because I'm deleting many sources at once?

Another thing is that I didn't check the "Delete media" check-box, but the text suggests that files are being deleted (they aren't actually being deleted). Maybe reconsider the phrasing here?

tcely:
Fixes #1603


meeb (approved):
Yep looks a lot more robust, thanks!

